PR #182, opened 2025-05-28
fix: add uv.lock and update docs

prasad-albert:


CONTRIBUTING.md
```diff
@@ -1,40 +1,43 @@
...
+pip install uv
 ```
 
 After that, the package and its dependencies can be installed
 in your local virtual environment by running
 
-```
-uv sync
+```bash
+uv sync --frozen
```
sfriedowitz, line 21: What is your motivation for adding uv lock to vcs? 

I would argue that since this is a library, not an application with mostly pinned dependencies, it is an unrealistic experience for developers to use frozen dependencies. End users will install the package from pypi, which will drift from this lock file.

sfriedowitz: The point I am arguing is that if we develop the package with frozen dependencies, but when a user installs the package there is some drift, we will be less likely to catch any bugs due to dependency drift while developing

sfriedowitz (approved):

CONTRIBUTING.md
prasad-albert: > What is your motivation for adding uv lock to vcs?
> 
> I would argue that since this is a library, not an application with mostly pinned dependencies, it is an unrealistic experience for developers to use frozen dependencies. End users will install the package from pypi, which will drift from this lock file.

okay i get your point now, we don't need the strict `--frozen`, i can fix that

PR #192, opened 2025-06-04
Add SheetFormulationRef object

jwelchalbert:
Adds a SheetFormulationRef resource to the Sheets class so that worksheets can deserialize the "Formulas" array. The Ref schema appears to be custom for the worksheet API as I couldn't find any other schemas that matched the fields exactly.

### DIFF
--- a/src/albert/resources/sheets.py
+++ b/src/albert/resources/sheets.py
@@ -294,7 +294,7 @@ class Sheet(BaseSessionResource):  # noqa:F811
 
     id: str = Field(alias="albertId")
     name: str
-    # formulations: list[Formulations] | None = Field(None)
+    formulations: list["SheetFormulationRef"] = Field(default_factory=list, alias="Formulas")
     hidden: bool
     _app_design: Design = PrivateAttr(default=None)
     _product_design: Design = PrivateAttr(default=None)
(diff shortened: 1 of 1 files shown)


src/albert/resources/sheets.py
```diff
@@ -294,7 +303,7 @@ class Sheet(BaseSessionResource):  # noqa:F811
 
     id: str = Field(alias="albertId")
     name: str
```
sfriedowitz, line 305: This is because the API returns unhydrated versions of the formula object? Can you use the formula object but leave its fields defaulting to Null/None?

jwelchalbert: Is that an InventoryItem? I don't see a base Formula object

sfriedowitz: @prasad-albert is going to pick up another ticket dealing with hydrated vs unhydrated objects from the listing/search endpoints. I think the pattern there is going to be use the same schema in the client, but allow it to be flexible such that it can be deserialized with default field values without additional API calls

sfriedowitz: If there is not an object already for this, then your implementation is fine

jwelchalbert: Ok sounds good I won't make any changes for now -- we can revisit later if needed.

sfriedowitz (approved):

PR #200, opened 2025-06-09
Feature/dc validations

lkubie:


src/albert/collections/data_templates.py
```diff
@@ -32,6 +46,13 @@ def create(self, *, data_template: DataTemplate) -> DataTemplate:
         DataTemplate
             The registered DataTemplate object with an ID.
         """
+        # Preprocess data_column_values to set validation to None if it is an empty list
+        # Handle a bug in the API where validation is an empty list
+        # https://support.albertinvent.com/hc/en-us/requests/9177
```
prasad-albert, line 51: I can't find this request,
image

src/albert/collections/data_templates.py
```diff
@@ -188,108 +209,38 @@ def update(self, *, data_template: DataTemplate) -> DataTemplate:
...
-                                "colId": old_dc_val.column_sequence,
-                            }
-                        )
-        if len(payload["data"]) > 0:
-            url = f"{self.base_path}/{existing.id}"
-            self.session.patch(url, json=payload)
-        return self.get_by_id(id=existing.id)  # always do this in case columns were added
+
+        path = f"{self.base_path}/{existing.id}"
+        # Handle special updates Imainly for complex validations
```
prasad-albert, line 217:
```suggestion
        # Handle special updates mainly for complex validations
```

lkubie:
@prasad-albert I updated where the patch utils lived and separated out the Objects and methods to avoid a circular dependency with the typing and imports.

src/albert/utils/patches.py
```diff
@@ -1,33 +1,412 @@
...
 
 from pydantic import Field
 
-from albert.utils.types import BaseAlbertModel
+from albert.resources.data_templates import DataTemplate
+from albert.resources.parameter_groups import (
+    EnumValidationValue,
+    ParameterGroup,
+    ValueValidation,
+)
```
sfriedowitz, line 10: Hmm, not sure I agree on the file location here. We have a bit of a circular module dependency now, which caused you to split the `utils/patches.py` into its own file to avoid circular imports.

This makes more sense being in `collections/patch_utils.py` to me, which can import `utils/patches.py` and `resources/parameter_groups.py`, etc.. without dealing with circularity.

src/albert/utils/patches.py
```diff
@@ -1,33 +1,412 @@
...
+from albert.utils.patch_types import PGPatchDatum
 
 
-class PatchOperation(str, Enum):
-    ADD = "add"
-    UPDATE = "update"
-    DELETE = "delete"
+class GeneralPatchDatum(PGPatchDatum):
```
sfriedowitz, line 14: Why is this one special and not part of `patch_types.py`?

src/albert/utils/patches.py
sfriedowitz: I think we can do this for now. Prasad proposed a module reorganization in an issue which will take care of some of these things

src/albert/utils/patches.py
lkubie: I missed it; good catch!

src/albert/utils/patches.py
prasad-albert: Yep, we can fix this later. It's good for now

src/albert/resources/tasks.py
```diff
@@ -49,6 +49,32 @@ class HistoryEntity(str, Enum):
     WORKFLOW = "workflow"
 
 
+class IntervalId(BaseAlbertModel):
+    id: str
+
+
+class BlockLevelInventoryInformation(BaseAlbertModel):
+    id: str | None = Field(default=None)
```
prasad-albert, line 57: we should alway have an `id` no?

src/albert/resources/tasks.py
```diff
@@ -49,6 +49,32 @@ class HistoryEntity(str, Enum):
...
+    lot_id: str | None = Field(default=None, alias="lotId")
+    inv_lot_unique_id: str | None = Field(default=None, alias="invLotUniqueId")
+
+
+class BlockState(BaseAlbertModel):
+    id: str | None = Field(default=None, description="The ID of the block.")
```
prasad-albert, line 63: same here ?
```suggestion
    id: str = Field(description="The ID of the block.")
```

lkubie: Same as above

PR #210, opened 2025-06-12
Update BT resources

sfriedowitz:


lkubie (approved):
LGTM! Bonus points if you add doc strings to your resources, too :)

PR #213, opened 2025-06-13
fix(dc, dt, links, lists, locations, lots): rename list

prasad-albert:


src/albert/collections/data_templates.py
```diff
@@ -251,3 +243,26 @@ def delete(self, *, id: str) -> None:
...
+
+    def get_all(
+        self,
+        *,
+        name: str | None = None,
+        user_id: str | None = None,
+        order_by: OrderBy = OrderBy.DESCENDING,
+        limit: int = 100,
+        offset: int = 0,
+    ) -> Iterator[DataTemplate]:
```
sfriedowitz, line 255: Since these kwargs are just dumped into `search`, do you think we can pass just a `**kwargs` and refer in the docstring to use the arguments there? Fine either way, just a suggestion

tests/collections/test_locations.py
```diff
@@ -4,7 +4,7 @@
 from albert.resources.locations import Location
 
 
-def _list_asserts(returned_list):
+def assert_location_items(returned_list):
     for i, c in enumerate(returned_list):
         if i == 30:
             break
```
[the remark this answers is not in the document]
prasad-albert: will fix this with max-items PR

src/albert/collections/data_templates.py
prasad-albert: ughh had to use kwargs for `projects` XD

src/albert/collections/data_templates.py
```diff
@@ -251,3 +244,38 @@ def delete(self, *, id: str) -> None:
...
+        def batched(iterable, size):
+            """Yield lists of up to `size` IDs from an iterable of objects with an `id` attribute."""
+
+            it = (item.id for item in iterable)
+            while batch := list(islice(it, size)):
+                yield batch
+
+        id_batches = batched(
+            self.search(name=name, user_id=user_id, order_by=order_by, limit=limit, offset=offset),
+            100,  # batch size
```
sfriedowitz, line 273: I think we should set the batch size equal to the `limit` keyword argument (soon to be renamed to `page_size`), such that every time we yield a page from the `search` API call we yield that entire batch as hydrated objects.

src/albert/collections/data_templates.py
```diff
@@ -251,3 +244,38 @@ def delete(self, *, id: str) -> None:
...
+
+        Use :meth:`search` for better performance.
+        """
+
+        def batched(iterable, size):
+            """Yield lists of up to `size` IDs from an iterable of objects with an `id` attribute."""
+
+            it = (item.id for item in iterable)
+            while batch := list(islice(it, size)):
+                yield batch
```
sfriedowitz, line 269: This is a pretty general method that would work on any iterator, so perhaps could be elevated to the `utils` module

Also, this has been implemented in `itertools` in Python 3.12 I believe, but sadly is above our min Python version :(

https://docs.python.org/3/library/itertools.html#itertools.batched

sfriedowitz: This is the dummy implementation given by `itertools`

```
def batched(iterable, n, *, strict=False):
    # batched('ABCDEFG', 3) → ABC DEF G
    if n < 1:
        raise ValueError('n must be at least one')
    iterator = iter(iterable)
    while batch := tuple(islice(iterator, n)):
        if strict and len(batch) != n:
            raise ValueError('batched(): incomplete batch')
        yield batch
```

prasad-albert: > This is a pretty general method that would work on any iterator, so perhaps could be elevated to the `utils` module
> 
> Also, this has been implemented in `itertools` in Python 3.12 I believe, but sadly is above our min Python version :(
> 
> https://docs.python.org/3/library/itertools.html#itertools.batched

yes, we could also create something like `CollectionMixin` for functions like these.

PR #215, opened 2025-06-17
Patch BTD references

sfriedowitz:


src/albert/resources/btdataset.py
```diff
@@ -10,7 +10,7 @@
 class BTDatasetReferences(BaseAlbertModel):
     project_ids: list[str]
     data_column_ids: list[str]
-    sheet_ids: list[str] = Field(default_factory=list)
+    sheet_ids: list[str] | None = Field(default=None)
```
sfriedowitz, line 13: Default should be None when not provided, rather than empty list

lkubie:
Post-approval haha

PR #220, opened 2025-06-22
refactor!: restructure SDK layout and standardize naming across modules

prasad-albert:


src/albert/core/auth/credentials.py
```diff
@@ -22,8 +22,8 @@ def from_env(
         *,
         client_id_env: str = "ALBERT_CLIENT_ID",
         client_secret_env: str = "ALBERT_CLIENT_SECRET",
-    ) -> Union["ClientCredentials", None]:
-        """Read `ClientCredentials` from the environment.
+    ) -> "AlbertClientCredentials":
```
sfriedowitz, line 25: Type hint is incorrect here. If the env vars are not set it returns `None`

src/albert/core/auth/credentials.py
```diff
@@ -102,6 +122,9 @@ def _requires_refresh(self) -> bool:
         )
 
     def get_access_token(self) -> str:
+        # TODO: enable when refresh token works
```
sfriedowitz, line 125: Were you able to figure out why this does not work? I gave up trying when first implementing the authentication flow.

sfriedowitz: Should we remove the stuff about refresh token then if it won't work for our auth flow?

src/albert/resources/common/patch.py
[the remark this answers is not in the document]
prasad-albert: i had some circular deps with it, so i decided to move it to core.

src/albert/core/auth/credentials.py
prasad-albert: i will remove it from this PR, will add in oauth PR

sfriedowitz: Yeah, let's come back to the auth flow options here. I think the abstraction can be improved a bit in the `TokenManager` class maybe, but I will have to see what you implement in the other PR first.

sfriedowitz (approved):

PR #222, opened 2025-06-23
refactor: disable deprecated list() method, enforce get_all usage

prasad-albert:
PS: review the second commit only


src/albert/collections/base.py
```diff
@@ -187,3 +189,13 @@ def _generate_patch_payload(
                         )
                     )
         return PatchPayload(data=data)
+
+    @deprecated(
+        "The 'list' method is deprecated and no longer supported. Use 'get_all(params=...)' or 'search(...)' instead.",
+        version="1.0.0",
```
sfriedowitz, line 195: This implies that we need to carry this around until 2.0? I think we can remove this entirely before then during our build up for 1.0 migration.

prasad-albert: this will go in with 1.0.0, and we can drop it in a couple of months?

PR #228, opened 2025-06-26
fix(data-templates): allow creation with no data columns

lkubie:


sfriedowitz (approved):

src/albert/collections/data_templates.py
```diff
@@ -174,7 +180,12 @@ def get_by_id(self, *, id: DataTemplateId) -> DataTemplate:
             The data template object on match or None
         """
         response = self.session.get(f"{self.base_path}/{id}")
-        return DataTemplate(**response.json())
+        try:
+            return DataTemplate(**response.json())
+        except Exception as e:
+            logger.warning(response.json())
```
prasad-albert, line 186: we don't log response usually, any specific reason here?

lkubie: Sorry this was me dubugging, will remove!

PR #237, opened 2025-07-11
Signed URL for notes, and get attachments by parents

sfriedowitz:


src/albert/collections/attachments.py
```diff
@@ -24,6 +24,51 @@ def _get_file_collection(self):
     def _get_note_collection(self):
         return NotesCollection(session=self.session)
 
+    def get_by_id(self, *, id: str) -> Attachment:
```
jwelchalbert, line 27: Is the ID here arbitrary or is it a standardized platform ID? if the latter can you create a new type?

src/albert/resources/tasks.py
```diff
@@ -92,6 +92,9 @@ class Standard(BaseAlbertModel):
     standard_organization: str | None = Field(
         alias="standardOrganization", default=None, frozen=True
     )
+    standard_organization_id: int | None = Field(
```
jwelchalbert, line 95: Same question as above -- is this a Company ID or an arbitrary str ID?

jwelchalbert (approved):

src/albert/collections/attachments.py
sfriedowitz: Good point, I will update this on my next working day

src/albert/resources/tasks.py
sfriedowitz: This is an arbitrary integer, which I do not think maps to an Albert ID.

We were missing fields on many of these schemas in the SDK, causing information required for the Python runner metadata to be lost during deserialization via the SDK

prasad-albert: Should we just forbid [extra](https://docs.pydantic.dev/2.0/usage/models/#extra-fields) fields to catch this schema drift faster?

sfriedowitz: In an ideal world, yes. But that will cause SOOOO many issues for us. I don't think its feasible right now.

prasad-albert: 🤔

sfriedowitz: Oh sorry, this is a different issue. There are a slew of other issues with the `Workflow` object in the SDK lol.

It looks like this field was just forgotten.

prasad-albert: > In an ideal world, yes. But that will cause SOOOO many issues for us. I don't think its feasible right now.
> 
> I suspect we will have to update most resources in the SDK.

I just realized that using extra = "forbid" is actually risky because it can break SDK users if the backend adds a new field, even if the SDK hasn't changed. Instead, we can have something like a `schema-drifter` running in CI to ensure our SDK models stay in sync with the backend schema by detecting any missing fields.

prasad-albert (approved):
LGTM, just tests missing.

src/albert/resources/tasks.py
sfriedowitz: Yeah, that would make more sense. Extra forbid in the client would cause lots of unexpected breaks.

I think an issue we will face with your proposal is that the backend OpenAPI docs are not always accurate. This is why we hand implemented the client schemas instead of using a generator

PR #242, opened 2025-07-15
release: v1.0.0 – features, refactors and breaking changes

prasad-albert:


src/albert/collections/custom_fields.py
```diff
@@ -160,8 +159,8 @@ def get_all(
         params = {
             "name": name,
             "service": service,
-            "lookupColumn": json.dumps(lookup_column) if lookup_column is not None else None,
-            "lookupRow": json.dumps(lookup_row) if lookup_row is not None else None,
+            "lookupColumn": lookup_column,
+            "lookupRow": lookup_row,
```
sfriedowitz, line 163: Does this work with our services? I think the way booleans were getting encoded by default was leading to validation issues from the API

src/albert/collections/data_columns.py
```diff
@@ -103,8 +102,8 @@ def deserialize(items: list[dict]) -> Iterator[DataColumn]:
             "orderBy": order_by.value,
             "startKey": start_key,
             "name": [name] if isinstance(name, str) else name,
-            "exactMatch": json.dumps(exact_match) if exact_match is not None else None,
-            "default": json.dumps(default) if default is not None else None,
+            "exactMatch": exact_match,
```
sfriedowitz, line 105: Same question, do we have test coverage the default boolean encoding works for this API?

src/albert/collections/custom_fields.py
prasad-albert: I moved the encoding logic to Paginator, my bad, it's in this commit -> https://github.com/albert-labs/albert-python/pull/242/commits/ede12841fc2e5227133064095e7d2d032405d774

src/albert/collections/data_columns.py
prasad-albert: We are still encoding the boolean, just moved it in paginator.

PR #252, opened 2025-07-23
fix(pagination): remove page_size from all collections

prasad-albert:


src/albert/core/pagination.py
```diff
@@ -75,14 +81,11 @@ def _create_iterator(self) -> Iterator[ItemType]:
             data = response.json()
             items = data.get("Items", [])
             item_count = len(items)
-            if not items:
-                return
 
-            yield from self.deserialize(items)
-
-            if item_count < self.page_size:
+            if not items and self.mode == PaginationMode.OFFSET:
```
prasad-albert, line 85: for DDB key-based pagination, there could be cases when `items=[]` but `lastKey` is still present in the response, which means there are still items in the next pages.

src/albert/collections/lists.py
```diff
@@ -101,11 +98,10 @@ def get_all(
         }
 
         return AlbertPaginator(
-            mode=PaginationMode.OFFSET,
```
sfriedowitz, line 104: Was this just broken before? How was this working with the incorrect page mode for the endpoint?

sfriedowitz (approved):
LGTM, but should wait until @ventura-rivera  comments on the CAS collection to merge

src/albert/collections/lists.py
prasad-albert: It was always broken. The tests were never fetching the next-page otherwise it would have raised error about `offset` as invalid query param.

src/albert/collections/cas.py
```diff
@@ -62,24 +60,32 @@ def get_all(
...
             session=self.session,
             params=params,
-            page_size=page_size,
-            max_items=max_items,
+            max_items=None,
             deserialize=lambda items: [Cas(**item) for item in items],
         )
 
+        # Apply custom filtering until https://linear.app/albert-invent/issue/TAS-564/inconsistent-cas-pagination-behaviour is fixed.
+        def filtered_items() -> Iterator[Cas]:
```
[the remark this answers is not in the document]
ventura-rivera: @prasad-albert this should be fine 🙂

ventura-rivera (approved):

PR #258, opened 2025-07-25
feat(custom_fields): allow customfields for CAS

prasad-albert:


ventura-rivera (commented):
Thanks @prasad-albert! I believe we need to add a new metadata field to the Cas class

ventura-rivera (approved):
lgtm!

PR #260, opened 2025-07-29
fix(auth): set token for specific tenant

prasad-albert:


src/albert/core/auth/sso.py
```diff
@@ -97,19 +105,40 @@ def refresh_token_url(self) -> str:
...
+        tenant_id = tenant_id.upper() if tenant_id else None
+
+        try:
+            token = (
+                next(t for t in tokens if t["tenantId"] == tenant_id) if tenant_id else tokens[0]
+            )
+        except (StopIteration, IndexError) as e:
+            raise AlbertAuthError("No matching tenant token found in response") from e
+
+        if not tenant_id:
```
lkubie, line 127:
```suggestion
        if not tenant_id and len(tokens) > 1:
```

lkubie (approved):
I see no need to warn if only one token but otherwise looks good!

PR #266, opened 2025-08-07
feat(sheets): support min/max in cells

prasad-albert:


src/albert/resources/custom_fields.py
```diff
@@ -110,7 +110,7 @@ class CustomField(BaseResource):
     required: bool | None = Field(default=None)
     multiselect: bool | None = Field(default=None)
     pattern: str | None = Field(default=None)
-    default: str | None = Field(default=None)
+    default: str | dict | None = Field(default=None)
```
prasad-albert, line 113: I have another PR incoming to strongly type this into pydantic model -- this is just to avoid breaking tests

PR #267, opened 2025-08-11
Ct task project migration updates

TomAlbertInvent:


src/albert/collections/custom_templates.py
```diff
@@ -87,6 +87,30 @@ def search(
...
+        -------
+        CustomTemplate
+            The created CustomTemplate object.
+        """
+
+        response = self.session.post(
+            url=self.base_path,
+            json=[
+                custom_template.model_dump(
+                    mode="json", by_alias=True, exclude_unset=True, exclude_none=True
```
prasad-albert, line 108: any reason why we explicitly exclude unset?

src/albert/resources/custom_templates.py
```diff
@@ -55,6 +56,8 @@ class GeneralData(BaseTaggedResource):
     priority: Priority | None = Field(default=None)
     sources: list[TaskSource] | None = Field(alias="Sources", default=None)
     parent_id: str | None = Field(alias="parentId", default=None)
+    notes: str | None = Field(default=None)
+    tags: list[Tag] | None = Field(default=None, alias="Tags")
```
prasad-albert, line 60: `GeneralData` inherits from `BaseTaggedResource` which has `tags` in it, so we don't need it here.

src/albert/resources/custom_templates.py
```diff
@@ -114,6 +117,8 @@ class BatchData(BaseTaggedResource):
     inventories: list[DataTemplateInventory] | None = Field(default=None, alias="Inventories")
     priority: Priority  # enum?!
     workflow: list[EntityLink] = Field(default=None, alias="Workflow")
+    notes: str | None = Field(default=None)
+    tags: list[Tag] | None = Field(default=None, alias="Tags")
```
prasad-albert, line 121: same here

src/albert/resources/custom_templates.py
```diff
@@ -126,6 +131,8 @@ class PropertyData(BaseTaggedResource):
     project: SerializeAsEntityLink[Project] | None = Field(alias="Project", default=None)
     inventories: list[DataTemplateInventory] | None = Field(default=None, alias="Inventories")
     due_date: str | None = Field(alias="dueDate", default=None)
+    tags: list[Tag] | None = Field(default=None, alias="Tags")
```
prasad-albert, line 134: same

src/albert/collections/custom_templates.py
```diff
@@ -87,6 +87,30 @@ def search(
             ],
         )
 
+    def create(self, *, custom_template: CustomTemplate) -> CustomTemplate:
```
prasad-albert, line 90: i think the API allows creation of multiple customtemplates at once, we could do something like this
```suggestion
    def create(self, *, custom_templates: list[CustomTemplate]) -> list[CustomTemplate]:
```

tests/collections/test_custom_templates.py
```diff
@@ -58,3 +58,31 @@ def test_hydrate_custom_template(client: Albert):
...
+
+    new_template = CustomTemplate(
+        name=seed_prefix,
+        category=seed.category,
+        data=(
+            seed.data.model_copy(update={"name": seed_prefix}, deep=True)
+            if getattr(seed, "data", None) is not None
+            else None
+        ),
+    )
```
prasad-albert, line 80: can we avoid seeding custom_templates, since we are not using it anywhere, and just define a model here instead, something like 

```suggestion
    new_template = CustomTemplate(
        name=seed_prefix,
        category=TemplateCategory.GENERAL,
        data=GeneralData(name=f"{seed_prefix}-general")
    )
```

PR #275, opened 2025-08-14
Fix/support legacy worksheets

jwelchalbert:
Adds a `default_factory` to the `sheets` field of the Worksheet model to support legacy worksheets which don't have multi sheet enabled.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -374,7 +374,7 @@ def seeded_worksheet(client: Albert, seeded_projects: list[Project]) -> Workshee
         wksht = collection.get_by_project_id(project_id=seeded_projects[0].id)
     except NotFoundError:
         wksht = collection.setup_worksheet(project_id=seeded_projects[0].id)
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -373,7 +373,7 @@ def seeded_worksheet(client: Albert, seeded_projects: list[Project]) -> Workshee
         wksht = collection.get_by_project_id(project_id=seeded_projects[0].id)
     except NotFoundError:
         wksht = collection.setup_worksheet(project_id=seeded_projects[0].id)
-    if wksht.sheets is None or wksht.sheets == []:
+    if not wksht.sheets:
```
jwelchalbert, line 376: This doesn't change the function of this conditional branching, it just eliminates the `None` check explicitly since it cannot ever be None now -- and this will still check for an empty list.

sfriedowitz (approved):
LGTM

PR #280, opened 2025-08-18
fix: custom field default type

prasad-albert:
# Summary

## Type (mark all that apply)

- [ ] Added — new feature
- [x] Fixed — bug fix
- [ ] Changed — existing functionality update
- [ ] Deprecated — will be removed soon
- [ ] Removed — feature removed

## Checklist

- [x] PR title follows Conventional Commits (e.g., `feat: add X`)
- [x] Tests added/updated for behavior changes
- [ ] Docs updated (if applicable)

### DIFF
--- a/src/albert/collections/custom_fields.py
+++ b/src/albert/collections/custom_fields.py
@@ -68,10 +68,10 @@ class CustomFieldCollection(BaseCollection):
         "min",
         "max",
         "entity_categories",
+        "default",
         # "required",
         # "multiselect",
         # "pattern",
-        # "default",
     }
     _api_version = "v3"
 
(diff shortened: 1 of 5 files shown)


prasad-albert:
> Should we have a proper data class for this? Something like:
> 
> ```python
> class DefaultValue(BaseAlbertModel):
>     field_type: FieldType = Field(alias="type")
>     value: ListItem
> ```

`value` does not appear to be a `ListItem`, it could be any of string, number or a list.
 i'm thinking to have concrete pydantic models, then have discriminated union alias -- something like,

```
DefaultValue = Annotated[
    StringDefaultValue | NumberDefaultValue | ListDefaultValue,
    Field(discriminator="type"),
]
```

ventura-rivera:
@prasad-albert that makes sense to me.

prasad-albert:
I did some testing and found issues with API validation and the response for `default`. I’ll hold off on this PR until I have more clarity from the API side. I don’t want to add “guessing” logic that could mask backend bugs.

src/albert/collections/custom_fields.py
```diff
@@ -237,6 +237,12 @@ def update(self, *, custom_field: CustomField) -> CustomField:
 
         # run patch
         url = f"{self.base_path}/{custom_field.id}"
-        self.session.patch(url, json=payload.model_dump(mode="json", by_alias=True))
+
+        self.session.patch(
+            url,
+            json=payload.model_dump(
+                mode="json", by_alias=True, exclude_unset=False, exclude_none=True
```
prasad-albert, line 244: `exclude_unset=False` because we also want to pass `type` from (StringDefault or NumberDefault or ListDefault)  which itself is a default Literal

Issue #287, opened 2025-09-10
Error: Update on Lots does not handle change in storage location well


Scriptalyst:
Hi all,
running an update on Lots, I encountered the following error, triggered by the part of the payload corresponding to StorageLocation:

BadRequestError: PATCH 'https://app.albertinvent.com/api/v3/lots/LOT408' failed with status code 400 (Bad Request). Errors: [{'label': 'data[0].attribute', 'msg': 'should be equal to one of the allowed values: storageLocation, manufacturerLotNumber, expirationDate, initialQuantity, inventoryOnHand, cost, status, packSize, hasNotes, hasAttachments, externalBarcodeId, Owner, workflowId', 'type': 'body'}, {'label': 'data[0].attribute', 'msg': 'format is invalid', 'type': 'body'}, {'msg': 'should match some schema in anyOf', 'type': 'body'}, {'label': 'data[0].attribute', 'msg': 'should be equal to one of the allowed values: barcodeId', 'type': 'body'}, {'msg': 'should match some schema in anyOf', 'type': 'body'}, {'msg': 'should match some schema in anyOf', 'type': 'body'}]
Body:
b'{"data": [{"operation": "update", "attribute": "StorageLocation", "newValue": {"status": "active", "Created": {"by": "USR35", "byName": "Moritz Haus", "at": "2025-08-19T07:25:09.684000Z"}, "name": "Mein erstes Lager", "albertId": "STL15", "Location": {"id": "LOC4"}}, "oldValue": {"id": "STL15"}}]}'

It seems like there should be no payload at all, since the StorageLocation is the same before and after (STL15). How do I come by the old and the new lot objects?
candidates = client.lots.get_all(inventory_id=lot.inventory_id)
        for candidate in candidates:
            if candidate.external_barcode_id.lower() == lot.external_barcode_id.lower():
                print(f"Lot {lot.external_barcode_id} already exists in Albert")
                # Update the lot... however, we filter the patch payload to avoid overwriting amounts
                payload = client.lots._generate_patch_payload(existing=candidate, updated=lot)
...

prasad-albert:
fixed in (https://github.com/albert-labs/albert-python/pull/292), v1.6.4

PR #292, opened 2025-09-22
fix(lots): update storage-location for lot

prasad-albert:


prasad-albert:
fixes: https://github.com/albert-labs/albert-python/issues/287

PR #298, opened 2025-09-25
feat(tasks): support task property data import

prasad-albert:


src/albert/resources/data_templates.py
```diff
@@ -22,15 +30,62 @@ class CSVMapping(BaseAlbertModel):
...
-    data_column_id: str = Field(alias="id", default=None)
+class AthenaMetadata(BaseAlbertModel):
+    table_name: str | None = Field(default=None, alias="tableName")
+    partition_key: str | None = Field(default=None, alias="partitionKey")
+
+
+class S3KeyReference(BaseAlbertModel):
+    rawfile: str | None = None
+    s3_input: str | None = Field(default=None, alias="s3Input")
+    s3_output: str | None = Field(default=None, alias="s3Output")
```
sfriedowitz, line 41: Can these be renamed to something more domain specific? I don't think we should be naming classes based on our AWS service names :)

src/albert/collections/proxy.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
prasad-albert: removed proxy collection

src/albert/utils/task_utils.py
```diff
@@ -0,0 +1,326 @@
...
+    return attachment_id
+
+
+def fetch_csv_table_rows(
+    *,
+    session: AlbertSession,
+    api_version: str,
+    attachment_id: str,
+    csv_table_key: str | None,
+) -> list[object]:
```
sfriedowitz, line 305: Is the type here really not known at all, even a `list[dict]` or `list[str]` or anything?

sfriedowitz (commented):
A few comments but looks good otherwise. Missing a minor version bump...

sfriedowitz (approved):

PR #324, opened 2025-11-12
Fix/update block workflow

mattkrain:
update_block_workflow() currently does not check if the existing_workflow_id is the "FINAL" workflow. 

**Issue 1:**
This causes the existing_workflow_id to become assigned to the "INITIAL" workflow, which can cause a patch error if the FINAL workflow does not equal the INITIAL workflow. 

**Issue 2:**
This also prevents the blocks FINAL workflow from being patched back to the INITIAL workflow


prasad-albert:
Superseded by #622 (merged in `58d5ca4b`). `main` now uses FINAL workflow id when updating block workflow, with INITIAL fallback and improved tests.

PR #363, opened 2026-02-05
feat: Added document search to projects collection

jwelchalbert:


albert-tb (approved):
looks good! Added two functions: document_search for retrieving all documents from a project, and upload_and_attach_document_to_project for uploading to the main platform

PR #374, opened 2026-02-18
feat: support targets api

sbernasek-albert:
- Add SDK bindings for the new /api/v3/targets CRUD API (create, get, list, delete)
- New Target resource model and TargetCollection registered on the Albert client as client.targets
- Integration tests and documentation pages

### DIFF
--- a/src/albert/collections/targets.py
+++ b/src/albert/collections/targets.py
@@ -28,8 +28,6 @@ class TargetCollection(BaseCollection):
         Retrieves a target by its ID.
     get_by_ids(ids) -> list[Target]
         Fetches multiple targets at once by their IDs.
-    list() -> list[Target]
-        Lists all active target entities.
     delete(id) -> None
         Deletes a target by its ID.
     """


src/albert/collections/targets.py
```diff
@@ -0,0 +1,121 @@
+from albert.collections.base import BaseCollection
+from albert.core.session import AlbertSession
+from albert.core.shared.identifiers import ProjectId
+from albert.core.utils import ensure_list
+from albert.resources.targets import Target
+
+
+class TargetCollection(BaseCollection):
```
sfriedowitz: Can you add something here indicating this is Alpha functionality and subject to breaking changes please?

@prasad-albert  do you have a preferred pattern for this?

prasad-albert: I say we do a pre-release to indicate this is still in beta, we could also add something like this in the docstring to indicate feature is unstable https://github.com/albert-labs/albert-python/blob/958fd3e85973330cad0e31f5fe4969a3f84da548/src/albert/collections/entity_types.py#L24

prasad-albert:
@sbernasek-albert  Can we change the PR title to - `feat: support targets api` and without exclamation, otherwise our release tool will treat it as breaking change and upgrade the major version instead of minor.

src/albert/collections/targets.py
```diff
@@ -0,0 +1,121 @@
...
+        Returns
+        -------
+        Target
+            The Target entity.
+        """
+        url = f"{self.base_path}/{id}"
+        response = self.session.get(url)
+        return Target(**response.json())
+
+    def list(self, *, project_id: ProjectId | None = None) -> list[Target]:
```
prasad-albert:
```suggestion
    def list(self, *, project_id: ProjectId | list[ProjectId] | None = None) -> list[Target]:
```
if they accept multiple ProjectIds

src/albert/__init__.py
```diff
@@ -4,4 +4,4 @@
 
 __all__ = ["Albert", "AlbertClientCredentials", "AlbertSSOClient"]
 
-__version__ = "1.17.0"
+__version__ = "1.18.0"
```
prasad-albert: We don't manually increment this anymore, it will be taken care by `release-please` tool

src/albert/resources/targets.py
```diff
@@ -0,0 +1,161 @@
...
+    EQ = "eq"
+    GT = "gt"
+    GTE = "gte"
+    LT = "lt"
+    LTE = "lte"
+    BETWEEN = "between"
+    IN_SET = "in-set"
+
+
+class ParameterCategory(str, Enum):
```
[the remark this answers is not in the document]
sbernasek-albert: I actually think their definition of a "parameter" entity is broken in the new targets api. I'm going to flag this to them so it will likely change in the next few days.

src/albert/collections/targets.py
sbernasek-albert: Added

prasad-albert (approved):

PR #465, opened 2026-04-09
fix(chats): allow clearing parentId by passing None to session update

prasad-albert:
## Summary
- `ChatSessionCollection.update()` previously treated `parent_id=None` as "not provided" and silently skipped it
- Introduced a `_UNSET` sentinel so callers can now pass `parent_id=None` to explicitly clear the folder (sends `{"newValue": null}` to the API)
- Removed `@validate_call` from `update()` to avoid pydantic coercing the sentinel at call time

### DIFF
--- a/src/albert/collections/chat_sessions.py
+++ b/src/albert/collections/chat_sessions.py
@@ -157,6 +157,7 @@ class ChatSessionCollection:
         ):
             yield session
 
+    @validate_call
     async def update(
         self,
         *,
--- a/src/albert/collections/chat_sessions.py
+++ b/src/albert/collections/chat_sessions.py
@@ -1,12 +1,15 @@
 from __future__ import annotations
 
 from collections.abc import AsyncIterator
+from typing import Any
 
 from pydantic import validate_call
 
 from albert.core.async_session import AsyncAlbertSession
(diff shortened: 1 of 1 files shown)


prakashchandrahv (approved):

prasad-albert:
> ## Code Review
> ### Issues Found
> * **[src/albert/collections/chat_sessions.py:157]** `@validate_call` was removed from `update()`. Per project standards, all public collection methods must have `@validate_call` for runtime validation. It was likely removed because pydantic rejects the `_UNSET` sentinel as a valid `str | None` value — but the fix is to use a typed sentinel (e.g. a dedicated `_UNSET` class with a proper type) rather than drop the decorator.
> * **[src/albert/collections/chat_sessions.py:162]** `parent_id: str | None = _UNSET` has a type mismatch: the declared type is `str | None` but the actual default is an `object()` sentinel. This is incorrect at the type level and would cause `validate_call` to reject calls that omit `parent_id` at runtime.
> 
> ### Summary
> 2 issues found. Removing `@validate_call` violates a hard project standard; the sentinel approach needs a proper typed sentinel (e.g. a singleton class) so the decorator can be restored.

@claude fixed both the issues -- do a review again

PR #467, opened 2026-04-09
feat(chats): add last_message_at field to ChatSession

prasad-albert:


prakashchandrahv (approved):

prasad-albert:
> ## Code Review
> ### Issues Found
> * **[src/albert/resources/chats.py:56]** `source_session_id` changed from `str | None` (with `default=None`) to required `str`. Any caller creating a `ChatSession` without providing this field will now get a Pydantic validation error at runtime. This is a breaking change — since breaking changes are not allowed in this repo, the field should either remain optional or the old default be preserved.
> * **[PR title]** Title prefix is `fix` but the PR adds new functionality (`last_message_at` field). Should be `feat(chats): add last_message_at field to ChatSession`.
> 
> ### Summary
> 2 issue(s) found. The breaking change to `source_session_id` is the critical one — it will silently break existing callers at runtime.


@claude shipping breaking change since this feature is marked Beta, and in a pre-release, so we should be good
change the PR title to reflect the new functionality

PR #488, opened 2026-05-11
feat: add support for get_data

sbernasek:


sbernasek:
> ## Code Review
> ### Issues Found
> * **[src/albert/resources/smart_datasets.py:138-143]** `SmartDatasetRecordIdentifier` fields `inventory_id`, `lot_id`, `task_id`, `property_data_id`, and `workflow_interval` are missing `Field(alias=...)` for their camelCase API counterparts (`inventoryId`, `lotId`, `taskId`, `propertyDataId`, `workflowInterval`). There is no `alias_generator` on `BaseAlbertModel`, so deserialization from the API response will silently leave these fields as `None` instead of raising an error. Every other resource in this codebase that maps snake_case to camelCase uses an explicit alias (see `lots.py`, `property_data.py`, `tasks.py`).
> * **[src/albert/resources/smart_datasets.py:188-213 / src/albert/collections/smart_datasets.py:45]** The `SmartDatasetData` docstring lists `identifiers : list[SmartDatasetIdentifier]` but the actual field type is `list[SmartDatasetRecordIdentifier]`. Separately, the class-level `Methods` docstring in `SmartDatasetCollection` lists `get_experiment_data(...)` but the actual method is `get_data(...)`.
> * **[src/albert/resources/smart_datasets.py:93]** `from_api_value` is an instance method (takes `self`) but ignores `self` entirely — it should be a `@classmethod` or `@staticmethod`. As written, calling `aggregate_by.from_api_value(data["aggregate_by"])` works by accident, but the signature is misleading and will confuse type checkers.
> 
> ### Summary
> 3 issues found. The missing camelCase aliases on `SmartDatasetRecordIdentifier` is the highest-severity issue — it will cause all identifier metadata to deserialize as `None` in production when the API returns camelCase keys.

@claude the API currently returns snakecase so point #1 is moot. The others are addressed

src/albert/collections/smart_datasets.py
```diff
@@ -183,3 +177,63 @@ def _generate_patch_payload(
...
+        """
+        Retrieves the experiment data for a smart dataset.
+
+        Parameters
+        ----------
+        id : SmartDatasetId
+            The ID of the smart dataset.
+        aggregate_by : SmartDatasetAggregateBy, optional
+            The aggregation level for the returned data. Defaults to ``ptd``.
+        ids : list[str], optional
```
prasad-albert, line 215: Are these also SmartDatasetIds?

prasad-albert (approved):

src/albert/collections/smart_datasets.py
sbernasek: No, these are Zeus record IDs. There is not albert object that can reliably define these.

PR #498, opened 2026-05-14
fix(data_templates): workaround backend s3key bug in calculation update test

prasad-albert:
## Summary
- Removes the session-scoped "Calculation Template" seed — no test referenced it after the rewrite
- Adds a function-scoped `calculation_dt` fixture in `test_data_templates.py` that owns the full lifecycle: create DT without calculation → add calculation via PATCH → yield → delete
- The two-step create+add approach is required because the backend retains `validation=[number]` on columns created with a calculation in the POST body, which triggers a backend 500 (`s3key` is not a valid attribute) on any subsequent calculation PATCH
- Adding the calculation via a separate PATCH causes the backend to clear `validation` to `[]`, after which further calculation updates succeed

### DIFF
--- a/tests/collections/test_data_templates.py
+++ b/tests/collections/test_data_templates.py
@@ -269,16 +269,43 @@ def test_enum_validation_update(client: Albert, seeded_data_templates: list[Data
         assert old in new_options
 
 
-def test_update_calculation(client: Albert, seeded_data_templates: list[DataTemplate]):
-    """Test updating calculation on a data template data column."""
-    dt = next(x for x in seeded_data_templates if "Calculation Template" in x.name)
-    column = next(x for x in dt.data_column_values if x.sequence != "COL0")
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


prasad-albert:
> ## Code Review
> ### Issues Found
> * **[tests/collections/test_data_templates.py:291]** The generator expression `next(c for c in dt.data_column_values if c.sequence not in ("COL0", "COL1"))` will always raise `StopIteration`. The fixture creates a DataTemplate with exactly 2 columns, which the backend assigns sequences `COL0` and `COL1`. The filter excludes both of those, so no element is ever found. The intent appears to be selecting the second column to attach a calculation to — the condition should be something like `c.sequence == "COL1"` (or `c.sequence not in ("COL0",)`).
> 
> ### Summary
> 1 issue found. The fixture will always crash before yielding, causing `test_update_calculation` to error on setup.

@claude matching on the exact DataColumn ID now, instead of iterating over the column names.

prakashchandrahv (approved):

PR #499, opened 2026-05-14
fix(tasks): add property blocks support to BatchTask

prasad-albert:
## Summary
- Add `blocks: list[Block] | None` field to `BatchTask` — the API has returned property blocks on batch task responses since the Connected Batch and Property Tasks GA release, but the SDK was silently dropping them
- Fix `update_block_workflow` to accept `BatchTask` in addition to `PropertyTask` (the hard `isinstance` guard was incorrectly restricting it)
- Update `add_block` and `remove_block` docstrings to reflect batch task support
- Seed a `BatchTask` with a property block and add three integration tests covering `add_block`, `update_block_workflow`, and `remove_block` on batch tasks

### DIFF
--- a/src/albert/collections/tasks.py
+++ b/src/albert/collections/tasks.py
@@ -104,7 +104,7 @@ class TaskCollection(BaseCollection):
     def add_block(
         self, *, task_id: TaskId, data_template_id: DataTemplateId, workflow_id: WorkflowId
     ) -> None:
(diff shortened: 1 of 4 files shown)


prakashchandrahv (approved):

ventura-rivera (approved):
LGTM!

PR #500, opened 2026-05-14
fix(sheets): scope recolor_column test to product design cells

prasad-albert:
## Summary
- `test_recolor_column` was asserting `c.color == CellColor.RED` on cells across all designs (product, apps, results), but the apps and results grid endpoints do not return `cellFormat` in `Values[]` for LKP-type columns
- The PATCH correctly stores the color server-side — this is a backend read gap in the grid endpoints, not an SDK bug
- Scoped the assertion to product design cells only, where `cellFormat` round-trips reliably for LKP columns
- Recolor still runs across all designs (correct behavior); only the verification is narrowed

### DIFF
--- a/tests/resources/test_sheets.py
+++ b/tests/resources/test_sheets.py
@@ -259,10 +259,13 @@ def test_add_formulation_no_clear_adds_new_column(
 
 
 def test_recolor_column(seeded_sheet: Sheet):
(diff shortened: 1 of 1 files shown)


siddhant-albertinvent (approved):
approved

PR #505, opened 2026-05-18
feat(cas): cap get_by_number partial-match pagination to avoid OpenSearch 500

prasad-albert:
## Summary
- `get_by_number(exact_match=False)` had no pagination cap, causing `CasPaginator` to drive `startKey` past 10000 — OpenSearch's deep-pagination hard limit — and return a 500 error
- Added `max_items: int | None = 50` (one page default) to `get_by_number` and `exists`, passed through to `get_all`; callers can override or pass `None` for unbounded search
- OpenSearch returns relevance-ranked results so the best match is always in the first page; deeper pagination is never necessary for these lookup methods

### DIFF
--- a/src/albert/collections/cas.py
+++ b/src/albert/collections/cas.py
@@ -137,7 +137,7 @@ class CasCollection(BaseCollection):
                 deserialize=lambda items: [Cas(**item) for item in items],
             )
 
-    def exists(self, *, number: str, exact_match: bool = True) -> bool:
+    def exists(self, *, number: str, exact_match: bool = True, max_items: int | None = 50) -> bool:
         """
         Checks if a CAS exists by its number.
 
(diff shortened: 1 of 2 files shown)


prasad-albert:
> ## Code Review
> ### Issues Found
> * **[src/albert/collections/cas.py:140]** `exists` is a public method modified by this PR but is missing `@validate_call`. The newly added `max_items` parameter won't be validated at runtime.
> * **[src/albert/collections/cas.py:247]** `get_by_number` is a public method modified by this PR but is missing `@validate_call`. Same issue — the new `max_items` parameter won't be validated.
> * **[PR title]** Title uses `fix:` but the change adds a new caller-visible parameter (`max_items`) to two public methods. Per AGENTS.md, new parameters are `feat`, not `fix`.
> 
> ### Summary
> 3 issues found. The missing `@validate_call` decorators are pre-existing but are on methods actively changed by this PR; the wrong commit type prefix means this change will be miscategorized in the changelog.

@claude Disagree on point 3. The max_items parameter was added solely to fix a deep-pagination bug that causes a 500 error, it's not a new capability being introduced.

prakashchandrahv (approved):

PR #514, opened 2026-05-20
fix(substance_v4): replace update_metadata diff params with keyword args

prasad-albert:
## Summary

- Remove the `current_metadata` / `updated_metadata` pair from `update_metadata` — callers no longer need to supply existing state
- Method now accepts individual keyword args per field; current state is fetched internally (consistent with `LotCollection.update`, `InventoryCollection.update`, etc.)
- Only the kwargs you pass are updated — unlisted fields are untouched; custom `metadata` dict follows the same rule (only listed keys affected, `None` value deletes a field)
- Promote `_UnsetType` / `_UNSET` sentinel from `chat_sessions.py` to `core/shared/types` so it can be reused across collections
- Adds `docs/examples/substance_v4.md` with examples for each metadata field type (string, single-select, multi-select, delete)

### DIFF
--- a/docs/examples/substance_v4.md
+++ b/docs/examples/substance_v4.md
@@ -1,0 +1,100 @@
+# Substances V4
+
+The `substances_v4` collection wraps the Albert v4 substance API, which exposes hazard data, custom tenant metadata, and structural identifiers at the CAS level.
+
+## Update metadata
+
+Use `update_metadata` to change specific fields on a tenant substance. Only the keyword arguments you pass are updated — everything else on the substance is left as-is.
+
+!!! example "Update scalar fields"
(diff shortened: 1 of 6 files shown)


prasad-albert:
> ### Mentions
> * PR title uses `fix:` but the change adds new caller-visible parameters (`notes`, `description`, `cas_smiles`, etc.) — per AGENTS.md, adding parameters is `feat:`.
> 
> ### Summary
> 1 bug found. The `getattr` without a default will raise `AttributeError` at runtime for any real substance whose scalar fields are absent from the API response.

@claude addressed both the issues, fixing a broken interface is fix not a feat.

PR #521, opened 2026-05-29
feat(sheets): added method for adding parameter group to worksheet process grid

sbernasek:


src/albert/resources/sheets.py
```diff
@@ -1197,7 +1253,7 @@ def delete_column(self, *, column_id: str) -> None:
             self.grid = None
 
     def delete_row(self, *, row_id: str, design_id: str) -> None:
-        endpoint = f"/api/v3/worksheet/design/{design_id}/rows"
+        endpoint = f"/api/v3/designs/{design_id}/rows"
         payload = [{"rowId": row_id}]
         self.session.delete(endpoint, json=payload)
```
prasad-albert, line 1259: row deletion is type-specific:
  ▎ - INV rows → DELETE /api/v3/worksheet/design/{id}/rows
  ▎ - PRG rows → DELETE /api/v3/designs/{id}/rows

Changing delete_row to always use /api/v3/designs/{id}/rows will break deletion of INV (and other non-PRG) rows.

prasad-albert (changes requested):

PR #523, opened 2026-06-02
feat: support smart projects

sbernasek:
## Summary

Adds first-class support for **smart projects** — the optimization-scope layer of a `Project` (a set of targets plus a built smart dataset).

- New `SmartProject` resource (`resources/smart_projects.py`): a session-bound `BaseSessionResource` reached lazily via the new `Project.smart` property.
- Smart operations live on `SmartProject` and mutate in place, returning `self`:
  - `add_target` — create a new target, or register an existing `Target`/`TargetId` to the scope.
  - `remove_target(delete=...)` — drop a target from the scope, optionally deactivating the target record.
  - `update_dataset(dataset=...)` — attach an existing dataset by ID, build from a given `SmartDatasetScope`, or build from the project's current scope when omitted.
  - `update` — low-level patch helper the above are built on.
- `ProjectCollection.create`/`get_by_id` now bind the session onto returned `Project`s so `project.smart` works off any fetched project.
- Integration tests in `tests/resources/test_smart_projects.py`.

### DIFF
--- a/src/albert/collections/projects.py
+++ b/src/albert/collections/projects.py
@@ -48,7 +48,7 @@ class ProjectCollection(BaseCollection):
         response = self.session.post(
             self.base_path, json=project.model_dump(by_alias=True, exclude_unset=True, mode="json")
         )
-        return Project(**response.json())
+        return Project(**response.json(), session=self.session)
 
     @validate_call
     def get_by_id(self, *, id: ProjectId) -> Project:
@@ -68,7 +68,7 @@ class ProjectCollection(BaseCollection):
         url = f"{self.base_path}/{id}"
         response = self.session.get(url)
 
-        return Project(**response.json())
+        return Project(**response.json(), session=self.session)
 
     def update(self, *, project: Project) -> Project:
         """Update a project.
--- a/src/albert/resources/projects.py
+++ b/src/albert/resources/projects.py
@@ -112,7 +112,13 @@ class Project(BaseSessionResource):
 
     @property
     def smart(self) -> SmartProject | None:
-        """Return the smart project resource for this project. If no smart project is found, returns None."""
+        """Return the smart project resource for this project.
+
+        Returns
+        -------
+        SmartProject or None
+            The smart project associated with this project, or None if no smart project exists.
+        """
         if self._smart is None:
             response = self.session.get(f"/api/v3/projects/{self.id}/getSmartProject")
             smart = response.json().get("smart", [])
--- a/src/albert/resources/projects.py
+++ b/src/albert/resources/projects.py
@@ -1,14 +1,17 @@
+from __future__ import annotations
+
 from enum import Enum
 
-from pydantic import Field, field_validator
+from pydantic import Field, PrivateAttr, field_validator
 
 from albert.core.base import BaseAlbertModel
 from albert.core.shared.identifiers import AttachmentId, ProjectId
-from albert.core.shared.models.base import BaseResource
+from albert.core.shared.models.base import BaseSessionResource
 from albert.core.shared.types import MetadataItem, SerializeAsEntityLink
 from albert.resources._mixins import HydrationMixin
 from albert.resources.acls import ACL
 from albert.resources.locations import Location
+from albert.resources.smart_projects import SmartProject
 
 
 class ProjectClass(str, Enum):
@@ -45,7 +48,7 @@ class GridDefault(str, Enum):
     WKS = "WKS"
 
 
-class Project(BaseResource):
+class Project(BaseSessionResource):
     """A project in Albert.
 
     Attributes
@@ -62,7 +65,6 @@ class Project(BaseResource):
         The metadata of the project. Optional. Metadata allowed values can be found using the Custom Fields API.
     prefix : str | None
         The prefix of the project. Optional.
-
     acl : list[ACL] | None
         The ACL of the project. Optional.
     task_config : list[TaskConfig] | None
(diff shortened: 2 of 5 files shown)


sbernasek:
> ## Code Review
> ### Issues Found
> * **[resources/projects.py:111]** `Project.smart` property docstring says _"converting the project to a smart project if it is not already"_ — that describes internal behaviour. AGENTS.md requires docstrings to describe what a method does from the caller's perspective, not internals. Also missing the Numpy-style `Returns` section that AGENTS.md requires for all new public methods.
> * **[resources/projects.py:75-76]** The class `Attributes` section documents `smart : SmartProject | None`, implying the property can return `None`. It cannot — the property raises `ValueError` when no smart project exists (line 121). Callers who write `if project.smart is None` will get an exception instead of `False`. Either remove `| None` from the attribute doc and clarify that `ValueError` is raised, or change the property to return `None` and update the return type annotation.
> * **[resources/smart_projects.py:188]** `scope.model_dump(by_alias=True, exclude_none=False, mode="json")` deviates from the project-wide convention of `exclude_none=True` and will send `null` for `sheet_ids` (explicitly set to `None` two lines above). If the backend requires the null for that field this is intentional, but it should be the explicit exception rather than a silent deviation.
> 
> ### Summary
> 3 issues found. The `smart` property's contradictory `| None` documentation is the highest-severity issue — it will silently mislead callers into writing incorrect null-checks that raise at runtime.

@claude Please check again - issues should be resolved by latest commit

prasad-albert (changes requested):
a few things to address,

  1. get_dataset missing — described in the PR body but not implemented.
  2. ProjectCollection.update() doesn't bind session — create and get_by_id were fixed, update was missed; the returned Project.smart will error.
  3. Unreachable guard — if self.scope is None in update_dataset can never be true since scope is a required field.
  4. test_get_smart is fragile — asserts isinstance(smart, SmartProject) but project.smart returns None if no smart project exists on a freshly seeded
  project.

  Design issues:

  - _smart stored as list[SmartProject] | None but only [0] is ever used — store as SmartProject | None directly.
  - API path /api/v3/projects/.../getSmartProject is hardcoded inside a resource model; should be a constant.

  Docs/style:

  - smart is a lazily-fetched @property but documented in the Attributes section as if it's a model field.
  - SmartProject class docstring is missing the required Methods section (per AGENTS.md).

sbernasek:
Regarding point 2 from @prasad-albert's review — **ProjectCollection.update() doesn't bind session**:

This is actually already handled. Tracing through the code:

- `update()` ([collections/projects.py:96](https://github.com/albert-labs/albert-python/blob/seb/smart-project-support/src/albert/collections/projects.py#L96)) returns `self.get_by_id(id=project.id)` as its final statement — it never constructs a `Project` from the PATCH response directly.
- `get_by_id()` was explicitly fixed in the same commit that fixed `create()` to pass `session=self.session` ([line 71](https://github.com/albert-labs/albert-python/blob/seb/smart-project-support/src/albert/collections/projects.py#L71)).

So `update()` was implicitly fixed when `get_by_id()` was fixed. The returned `Project` always has `_session` set, and `Project.smart` will work correctly. There are only two places in the collection that construct and return a `Project` object (lines 51 and 71), and both pass `session=self.session`.

The review saw `create` and `get_by_id` explicitly patched and flagged `update` as missed — but `update` doesn't construct its own `Project`; it delegates to the now-fixed `get_by_id`.

@claude Can you confirm this assessment? Specifically: does `ProjectCollection.update()` return a session-bound `Project` given the current code, and would `Project.smart` work on the result?

sbernasek:
Addressed the remaining items from the review. Here's what changed:

### Unreachable guard removed
`if self.scope is None` in `SmartProject.update_dataset` was removed. `scope` is a required field with no default so this branch could never execute.

### Hardcoded API path extracted to a constant
`/api/v3/projects` was appearing as a magic string in both `resources/projects.py` (inside the `smart` property) and in every method of `resources/smart_projects.py`. Extracted to `_PROJECTS_BASE_PATH = "/api/v3/projects"` at the top of `smart_projects.py` and imported from there into `projects.py`.

### `smart` removed from `Project.Attributes` docstring
`smart` is a lazily-fetched `@property`, not a Pydantic model field. It was documented in the `Attributes` section as if it were a field. Removed it from `Attributes` — the property carries its own Numpy-style docstring.

### `SmartProject` class docstring — `Methods` section added
Added the required `Methods` section listing `add_target`, `remove_target`, and `update_dataset` per AGENTS.md.

### `test_get_smart` hardened
The test was asserting `isinstance(smart, SmartProject)` unconditionally, but freshly seeded projects have no smart project so `project.smart` returns `None`. Updated to `assert smart is None or isinstance(smart, SmartProject)` with the `project_id` check gated on non-None.

---

Regarding **`get_dataset` missing** (point 1): intentionally not implemented. The resource already exposes `dataset_id`; callers who want to operate on the dataset can fetch it via `client.smart_datasets.get_by_id(id=smart.dataset_id)`. Adding a convenience getter here would duplicate the collection's responsibility.

Regarding **`_smart` stored as `list`** (design note): intentional. The underlying API schema returns `smart` as a list (always with at most one entry). Keeping it as `list[SmartProject] | None` in the private attribute avoids a custom deserializer; the `@property` wraps it into the intuitive `SmartProject | None` interface for callers.

Tests: `uv run pytest tests/resources/test_smart_projects.py tests/collections/test_projects.py` — all 13 relevant tests pass.

sbernasek:
@claude please review the above commit and make sure it addresses Prasad's comments

prasad-albert (approved):
LGTM

PR #528, opened 2026-06-08
feat(sheets): row grouping, new column/row types, and fixes from PR #267

prasad-albert:
## Summary
- **fix(custom_templates):** add missing `notes`, `due_date`, `NotebookData.id` fields; fix ACL discriminated union to accept backend tokens (`CustomTemplateTeam` etc.); fix tag resolution to fall back to nested `Data.Tags`
- **fix(workflows):** `create()` now handles `existingAlbertId` dedup response from the bulk endpoint; `ParameterGroups` defaults to `[]` since the API never returns them on create
- **feat(sheets):** row grouping (`Design.group_rows`, `Design.get_groups`), `RowConfig`/`RowGroup` models, new `Row` hierarchy fields, `ColumnPosition`/`RowPosition` enums, new column methods (`add_blank_column`, `add_lookup_column`, `add_function_column`, `add_property_column` — all backed by `_add_column`), new row methods (`add_lookup_row`, `add_app_row`), column state methods (`pin_columns`, `unpin_columns`, `set_columns_width`, `hide_column`, `show_column`), `add_components_to_formulation`, and `CellType.FNC`

### DIFF
--- a/src/albert/collections/workflows.py
+++ b/src/albert/collections/workflows.py
@@ -31,8 +31,13 @@ class WorkflowCollection(BaseCollection):
         self.base_path = f"/api/{WorkflowCollection._api_version}/workflows"
 
     def create(self, *, workflows: list[Workflow]) -> list[Workflow]:
-        """Create or return matching workflows for the provided list of workflows.
-        This endpoint automatically tries to find an existing workflow with the same parameter setpoints, and will either return the existing workflow or create a new one.
+        """Create workflows, or return existing ones if a match is found.
+
+        Deduplicates by parameter setpoints — if a workflow with identical
+        setpoints already exists, it is returned instead of creating a new one.
+        Returned workflows will have an empty ``parameter_group_setpoints`` list
+        regardless of whether they were created or matched; use ``get_by_id`` to
+        fetch the full setpoints if needed.
 
         Parameters
         ----------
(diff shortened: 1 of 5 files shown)


prasad-albert:
> ## Code Review
> ### Issues Found
> * **[sheets.py:482]** `config.model_dump(exclude_none=True)` is missing `by_alias=True` and `mode="json"`. Should be `config.model_dump(by_alias=True, mode="json", exclude_none=True)` to follow the project-wide serialization convention (AGENTS.md). Harmless now since `RowConfig` has no aliases and only `str` fields, but will silently break if the model ever grows an aliased or non-primitive field.
> * **[sheets.py:266–268]** `get_groups` manually checks `response.status_code >= 400` and returns an empty list. Every other `session.get/put/post/patch` call in this file lets the SDK raise on HTTP errors — no status-code guards anywhere else. If the session auto-raises on 4xx/5xx (the normal pattern), this guard is dead code. If it doesn't, it silently discards all errors — permissions failures, server errors — and callers see an empty list with no indication something went wrong. The correct fix is either to drop the guard (let the SDK raise) or catch a specific `AlbertHTTPError` subclass only for the expected 404 case.
> * **[sheets.py]** `add_components_to_formulation` is a public method on `Sheet` but lacks `@validate_call`, inconsistent with every other new public `Sheet` method added in this same PR (`add_lookup_row`, `add_app_row`, `add_blank_column`, `add_lookup_column`, `add_function_column`, `add_property_column`, `pin_columns`, `unpin_columns`, `set_columns_width`, `hide_column`, `show_column`).
> 
> ### Summary
> 3 issues found. The HTTP error handling in `get_groups` is the most impactful — it either silently swallows real errors or is dead code depending on session behavior; the `model_dump` and missing decorator are lower-risk pattern violations.

@claude re-review the PR, the comments have been addressed

mattkrain (approved):
Looks good!

PR #529, opened 2026-06-08
feat(tasks): support updating project via tasks.update()

prasad-albert:
## Summary
- `tasks.update()` previously had no support for changing a task's associated project
- Adds `project` as a patchable field in `generate_adv_patch_payload`, handling all three cases: add (no existing project), update (change project), and delete (remove project)
- Uses the correct API operations: `add` / `update` / `delete` on `attribute: projectId`

### DIFF
--- a/src/albert/collections/tasks.py
+++ b/src/albert/collections/tasks.py
@@ -723,7 +723,7 @@ class TaskCollection(BaseCollection):
 
         Notes
         -----
-        The following fields can be updated: ``due_date``, ``metadata``, ``name``, ``priority``, ``state``.
+        The following fields can be updated: ``due_date``, ``metadata``, ``name``, ``priority``, ``project``, ``state``.
         """
         existing = self.get_by_id(id=task.id)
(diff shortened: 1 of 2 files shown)


prasad-albert:
> ### Issues Found
> * **[src/albert/utils/tasks.py:387-417]** The `project` field on `BaseTask` is typed as `SerializeAsEntityLink[Project] | list[SerializeAsEntityLink[Project]] | None`. For a `BatchTask` whose `project` is a list, `getattr(old_project, "id", None)` returns `None` (lists have no `.id`), so `old_id` and `new_id` both resolve to `None`, the `new_id == old_id` guard fires, and the update is silently dropped. If only one side is a list (e.g. old is a list, new is a single project), the code fires an incorrect ADD instead of UPDATE. The implementation should guard against the list case — either raise `ValueError` for unsupported input, or branch on `isinstance`.

@claude fixed this by guarding against it

lkubie (approved):
LGTM

PR #539, opened 2026-06-10
feat(targets): widen TargetParameter.value to operator/value-pair with legacy coercion

sbernasek:
## What

`TargetParameter.value` now accepts the full operator/value-pair shape (`{operator, value}`) introduced by the api-targets contract update, instead of a bare scalar. A `ParameterFilter = TargetValue` alias makes the intent explicit in code. Bare legacy scalars are tolerated on read: numeric → `{operator: eq}`, string → `{operator: in-set, value: [s]}`.

## Why

The api-targets service is migrating parameter filter storage from bare scalars to operator/value-pair objects (branch `anubhav/aim-766`). Backfill is lazy, so clients can still encounter bare scalars for targets not yet read since deploy. This change (ML-1218) ships the SDK side independently of the api-targets deploy, making it safe to upgrade before or after.

## How

- `TargetValue` and `TargetRangeValue` are now defined *before* `TargetParameter` in the file so there is no forward-reference issue with the validator.
- A Pydantic `field_validator("value", mode="before")` handles coercion at deserialization time, mirroring the server-side `coerceLegacyParameterFilter` logic exactly.
- `bool` is explicitly checked before `int` to prevent `True`/`False` from becoming `eq` numeric filters.
- `ParameterFilter = TargetValue` is an alias (not a subclass) — if we need to restrict operators to the contract subset later we can promote it, but DRY wins for now.
- The existing `model_dump(by_alias=True, mode="json", exclude_none=True)` serialization path already handles nested `TargetValue` correctly — no collection-layer changes needed.

## Testing

Run the model-validation tests immediately (no API required):

```bash
uv run pytest tests/collections/test_targets.py::TestTargetParameterCoercion -v
```

Seven coercion cases pass: numeric→eq, int→eq, string→in-set, bool→in-set (not eq), None passthrough, structured dict passthrough, and in-set list passthrough.

The integration test `test_target_create_with_between_parameter_filter` exercises a create+read round-trip against the dev API once api-targets (`anubhav/aim-766`) is deployed.

Resolves ML-1218. Part of ML-1207.


src/albert/resources/targets.py
```diff
@@ -117,6 +89,61 @@ class TargetValue(BaseAlbertModel):
     value: TargetRangeValue | str | float | list
 
 
+# Parameter filters use the same operator/value-pair schema as target values.
+ParameterValue = TargetValue
```
sumedhb1995: Why are we creating a new variable that has the exact same value as the old variable? Shoudln't we just use that directly?

tests/seeding.py
```diff
@@ -1884,6 +1892,41 @@ def generate_target_seeds(
         ),
     ]
 
+    if seeded_parameters is not None:
+        params_template = next(
+            (
+                x
+                for x in seeded_data_templates
+                if x.name == f"{seed_prefix} - Parameters Data Template"
```
sumedhb1995, line 1902: Do we want to use a constant here instead of a magic string for filtering?

sumedhb1995 (approved):
Overall looks good, although I have little context in this area. A few comments from a general production readiness perspective that would be good to address prior to merging

src/albert/resources/targets.py
sbernasek: just for aliasing reasons. It's a parameter value not a target value but they are structurally identical. we cant rename it to some common name like "FilterValue" because then it would be inconsistent with the API contract

tests/seeding.py
sbernasek: The prefix is set by a uuid4() call in the pytest fixture (used all over the SDK) so I'm not sure how we would know the expected constant value

prasad-albert (approved):
LGTM, the tests failed with a weird error, re-running them

PR #543, opened 2026-06-12
feat(parameter_groups): support User-type special parameter values

prasad-albert:
## Summary
- Adds `User` to the `ParameterValue.value` union type (`str | InventoryItem | User | None`) to support the upcoming User-type special parameter feature
- When a parameter group contains a User-type special parameter (value ID prefixed `USR...`), the SDK now deserialises it as a `User` entity link rather than falling back to a bare `EntityLink`
- No breaking changes — existing INV and TMP special parameters are unaffected
- Matches the API contract documented in the backend solution spec (no new fields; same `{id, name}` wire shape)


lkubie (approved):
LGTM

PR #549, opened 2026-06-15
feat(substance_v4, attributes): promote substance v4 and attributes to stable

prasad-albert:
## Summary
- Cherry-picks `SubstanceCollectionV4` (v4 substance API) from the `next` beta branch into `main` for a stable release
- Includes the `update_metadata` keyword-args fix (`fix(substance_v4)`) from v1.26.1-beta0
- Cherry-picks `AttributeCollection` with full CRUD and reference-value support from v1.27.0-beta0
- All 22 integration tests pass (5 substance_v4, 17 attributes)


prasad-albert:
> ## Code Review
> ### Issues Found
> * **[src/albert/resources/substance_v4.py:2326]** `SubstanceV4Create.metadata` has `alias="Metadata"` (capital M), but `SubstanceV4Metadata.metadata` has no alias (serializes as `"metadata"` lowercase). `create()` calls `model_dump(by_alias=True)`, so the field is sent as `"Metadata"` in the POST body. If the API expects lowercase `"metadata"`, any `create()` call that includes metadata will silently send the wrong key and the metadata will be ignored. The alias should either match the wire format used by the backend or be removed to serialize as `"metadata"`.
> 
> ### Summary
> 1 issue found. Silent data loss in `SubstanceV4Collection.create()` when `metadata` is supplied.

confirmed with the API spec, this is the correct implementation and alias.

prasad-albert:
> > ## Code Review
> > ### Issues Found
> > 
> > * **[src/albert/resources/substance_v4.py:2326]** `SubstanceV4Create.metadata` has `alias="Metadata"` (capital M), but `SubstanceV4Metadata.metadata` has no alias (serializes as `"metadata"` lowercase). `create()` calls `model_dump(by_alias=True)`, so the field is sent as `"Metadata"` in the POST body. If the API expects lowercase `"metadata"`, any `create()` call that includes metadata will silently send the wrong key and the metadata will be ignored. The alias should either match the wire format used by the backend or be removed to serialize as `"metadata"`.
> > 
> > ### Summary
> > 1 issue found. Silent data loss in `SubstanceV4Collection.create()` when `metadata` is supplied.
> 
@claude confirmed with the API spec, this is the correct implementation and alias.

src/albert/resources/substance_v4.py
```diff
@@ -0,0 +1,276 @@
...
+    value: str
+
+
+class SubstanceV4Attribute(BaseAlbertModel):
+    """An attribute entry for creating a substance.
+
+    Attributes
+    ----------
+    attribute_name : str
+        The attribute name (e.g. ``hazards``, ``name``).
```
lkubie:
```suggestion
        The attribute name (e.g. ``hazards``, ``name``). Note that this name must exactly match one of the allowed Substance Attribute Names, and in the future will reference an attribute ID
```

@prasad-albert Kaushal is the best person to confirm with here

lkubie (changes requested):

src/albert/resources/substance_v4.py
KaushalKishoreKK: This contract is correct for now, it will change in future once we define the global attributes for substance

prasad-albert:
> Code Review
> 
> Issues Found
> 
> * **src/albert/collections/substance_v4.py:428** — In `update_metadata`, the current value for the scalar fields is read via `getattr(substance, attr, None)` using snake_case names (`cas_smiles`, `inchi_key`, `iupac_name`, `cactus_status`). `SubstanceV4Info` does not declare these fields; it relies on `extra="allow"`, which stores them under their camelCase wire keys (`casSmiles`, `inchiKey`, `iUpacName`, `cactusStatus`). So `getattr(..., "cas_smiles", None)` always returns None, and these four fields always emit an `add` patch operation even when the substance already has a value, instead of the `update` (with `oldValue`) branch. `notes`/`description` work only because their wire name equals the attribute name. Updating an existing SMILES/InChIKey/IUPAC/CACTUS value therefore sends the wrong patch op (a diff-based PATCH endpoint typically requires `update` + `oldValue` when a value exists). Declare these fields on `SubstanceV4Info` (or read them via the camelCase alias) so the old value is resolved.
> 
> Summary
> 
> 1 issue found. Latent correctness bug on the `update_metadata` update path for camelCase-aliased scalar fields; the create path and integration test only exercise the fresh-substance (add) case, so it is not caught.

@claude this is fixed now -- re-review the PR

lkubie (approved):

PR #556, opened 2026-06-18
feat(chats): add pageContext to ChatMessage (AI-637)

sumedhb1995:
## What

Chat messages can now carry an optional `pageContext` map — the embed page context the user was on when they sent the message. This lets api-chat persist and return that context on chat reload.

## Why

Phase 2 of context-aware Ask Albert needs the SDK to round-trip `pageContext` between the ReAct worker and api-chat. Without this field on `ChatMessage`, the worker cannot write context to storage or read it back when rebuilding conversation history ([AI-637](https://linear.app/albert-invent/issue/AI-637/accept-metadatapage-context-on-search-requests)).

## How

- Add `page_context: dict[str, Any] | None` with `pageContext` alias on `ChatMessage`.
- Unit test serialization round-trip (`pageContext` ↔ `page_context`).
- Serialization and deserialization test

## Testing

```bash
uv run pytest tests/resources/test_chats.py -q
uv run pytest tests/collections/test_chat_messages.py -q  # live api-chat
```

## Merge order

Merge alongside or before the api-chat schema PR. The albert-ai persistence PR depends on this SDK field.


tests/collections/test_chat_messages.py
```diff
@@ -0,0 +1,43 @@
...
+from albert.resources.chats import (
+    ChatComponentType,
+    ChatMessage,
+    ChatRole,
+    ChatUserType,
+)
+
+
+@pytest.mark.asyncio
+async def test_chat_message_page_context_round_trip(async_client, seeded_session) -> None:
```
prasad-albert: Could you add type hints here?

tests/collections/test_chat_messages.py
```diff
@@ -0,0 +1,43 @@
...
+    """Test page_context persists on create and reloads on get_by_id."""
+    page_context = {
+        "url": "https://app.albertinvent.com/#notebook",
+        "entity": "notebook",
+        "albert_id": "NTB1",
```
prasad-albert: Is this a real seeded notebook? will it always be available?

tests/collections/test_chat_messages.py
```diff
@@ -0,0 +1,43 @@
...
+    ChatRole,
+    ChatUserType,
+)
+
+
+@pytest.mark.asyncio
+async def test_chat_message_page_context_round_trip(async_client, seeded_session) -> None:
+    """Test page_context persists on create and reloads on get_by_id."""
+    page_context = {
+        "url": "https://app.albertinvent.com/#notebook",
```
prasad-albert: Can we use `ALBERT_BASE_URL` here?

src/albert/resources/chats.py
```diff
@@ -4,6 +4,7 @@
 from typing import Any
 
 from pydantic import Field
+from typing_extensions import TypedDict
```
prasad-albert: Since we target py3.10>= we can just import from `typing` here

tests/conftest.py
```diff
@@ -87,6 +89,28 @@
 )
 from tests.utils.fake_session import FakeAlbertSession
 
+_BOT_EXTERNAL_USER_ID_HEADER = "x-external-user-id"
+
+
+def _bot_external_user_id(sync_client: Albert) -> str:
+    if external_user_id := os.getenv("ALBERT_EXTERNAL_USER_ID_SDK"):
```
prasad-albert: What is this new env var for? Can you add documentation around this?

tests/conftest.py
```diff
@@ -87,6 +89,28 @@
...
+    return str(sync_client.users.get_current_user().id)
+
+
+def _configure_async_client_for_bot(async_client: AsyncAlbert, sync_client: Albert) -> None:
+    """Set x-external-user-id for bot OAuth tokens (required by chat APIs)."""
+    auth_manager = async_client.session._auth_manager
+    if auth_manager is None:
+        return
+    token = auth_manager.get_access_token()
+    payload = jwt.decode(token, options={"verify_signature": False, "verify_aud": False})
```
prasad-albert: This exact pattern of decoding JWT re-introduces security vuln flagged previously. This is a blocker -- we should find an alternative solution for this

prasad-albert (changes requested):

src/albert/resources/chats.py
sumedhb1995: The change you are recommending fails CI unfortunately: https://circleci.com/gh/albert-labs/albert-python/7256

sumedhb1995:
@prasad-albert I have updated the test to be a simpler, non-integration test to avoid the issues you highlighted and not require a framework level change. As for the one comment on business logic around `typing_extensions`, CI looks like its still running on 3.10 so I need to keep `typing_extensions.TypedDict`

src/albert/resources/chats.py
prasad-albert: okay wasn't aware of this limitation 👍

prasad-albert (approved):

PR #559, opened 2026-06-19
fix(units): patch synonyms as item-level add/delete operations

prasad-albert:
## Summary
- Updating a unit's `synonyms` previously sent a single whole-list `update` op, which the backend rejects — it has no `update` operation for `Synonyms` (only item-level `add`/`delete`, each with a scalar value), and it rejects more than one operation on the same attribute in a single request.
- `synonyms` changes are now diffed into per-item `add`/`delete` operations, and each `Synonyms` operation is sent as its own PATCH request (other attributes like `symbol`/`category` still go in one request).
- Folded synonyms coverage into the existing `test_unit_crud` (no new fixture/seeding) — adds and removes a synonym and asserts the result. Verified end-to-end against `app.albertinvent.com` (full `test_units.py` suite: 8 passed).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,7 +67,7 @@ ipython_config.py
 __pypackages__/
 
 # Environments
-.env
+.env*
 .venv
 env/
 venv/
--- a/src/albert/collections/units.py
+++ b/src/albert/collections/units.py
@@ -8,6 +8,7 @@ from albert.core.pagination import AlbertPaginator
 from albert.core.session import AlbertSession
 from albert.core.shared.enums import OrderBy, PaginationMode
 from albert.core.shared.identifiers import UnitId
+from albert.core.shared.models.patch import PatchDatum, PatchOperation, PatchPayload
 from albert.core.utils import ensure_list
 from albert.resources.units import Unit, UnitCategory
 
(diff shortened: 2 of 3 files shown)


m-j-hofmann (approved):

prasad-albert (commented):
**Unset synonyms will delete all existing synonyms**

`_generate_unit_patch_payload` reads `updated.synonyms or []` without checking `model_fields_set`. If a caller builds a partial update object without setting `synonyms`:

```python
unit = Unit(id="x", symbol="y")   # synonyms never set → not in model_fields_set
client.units.update(unit=unit)
```

`updated.synonyms or []` resolves to `[]`, so `existing_synonyms - []` emits a `DELETE` for every existing synonym. Same unset-vs-empty bug fixed in `base.py` / `generate_adv_patch_payload` in #561.

One-liner fix — add this right after stripping the old Synonyms datum:

```python
payload.data = [d for d in payload.data if d.attribute != "Synonyms"]
if "synonyms" not in updated.model_fields_set:
    return payload   # caller never set synonyms → leave them untouched
```

Explicit `synonyms=[]` still lands in `model_fields_set` so intentional clears are preserved.

PR #560, opened 2026-06-23
fix(exceptions): make AlbertHTTPError picklable for Ray/multiprocessing contexts

sbernasek:
## What

`AlbertHTTPError` and all its subclasses now survive a `pickle` round-trip. The exception's message is preserved; `response` is set to `None` after unpickling (since `requests.Response` is not serializable).

## Why

The breakthrough/Zeus backend dispatches work to remote Ray workers. Ray serializes exceptions across process boundaries using pickle. `AlbertHTTPError.__init__` takes a `requests.Response` object, but Python's default exception pickling stores `args=(message,)` and calls `__init__(message)` on reconstruction -- causing `AttributeError: 'str' object has no attribute 'json'`. This surfaced as a `RaySystemError: Failed to unpickle serialized exception` in CI, masking the real error and blocking recovery in integration tests.

## How

- Added `__reduce__` to `AlbertHTTPError` returning `(_restore_albert_http_error, (type(self), self.message))`. This bypasses `__init__` entirely on unpickle.
- `_restore_albert_http_error` uses `Exception.__new__(cls, message)` to construct the instance without calling `__init__` -- the same pattern already used in `handle_async_http_errors` in the same file.
- All subclasses inherit `__reduce__` and are covered automatically.

## Testing

10 new unit tests in `tests/test_exceptions.py`:

```bash
uv run pytest tests/test_exceptions.py -v
```

Parametrized over all 8 `AlbertHTTPError` subclasses -- each is raised with a realistic 404 response and round-tripped through `pickle.dumps`/`pickle.loads`. Tests verify type is preserved, message content survives, and `response` is `None` after unpickling.

### DIFF
--- a/src/albert/exceptions.py
+++ b/src/albert/exceptions.py
@@ -17,6 +17,20 @@ class AlbertAuthError(AlbertException):
     """Raised when authentication fails (e.g., bad credentials, expired token)."""
 
 
+def _restore_albert_http_error(cls: type, message: str) -> "AlbertHTTPError":
+    """Reconstruct an AlbertHTTPError from a pickled message string.
+
+    Python's default exception pickling stores args and calls __init__(*args)
+    on unpickle. AlbertHTTPError.__init__ expects a requests.Response, not a
+    string, so the default path fails. This function bypasses __init__ and
+    reconstructs the exception from the pre-formatted message alone.
+    """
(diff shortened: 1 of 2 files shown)


xperrylinn (approved):
LGTM

PR #561, opened 2026-06-24
fix(patch): leave unset fields untouched in update payloads

prasad-albert:
## Problem

Patch-payload generators decided `delete` ops from `value is None` alone. A field the caller never set defaults to `None` (or is coerced to `[]` by `getattr(updated, attr) or []` in the task special-attr loop), so omitted fields were read as deletions.

For partial update objects (the dominant Ask Albert agent pattern), this emitted bad `delete` ops that either:
- the API rejects (e.g. `"Delete operation not allowed for attribute-name"`), or
- silently wipe existing data (inventory rows, tags).

## Fix

Gate the diff on `attribute in updated.model_fields_set`:

- **unset** (`not in model_fields_set`) -> no-op, leave server value
- **explicit `None`** -> delete (unchanged)
- **explicit `[]` / `{}`** -> clear to empty (unchanged)

Non-breaking: the fetch-mutate-update pattern is unchanged (fetched objects have all server fields in `model_fields_set`); only erroneous deletes of omitted fields stop. Explicit clears still work.

### DIFF
--- a/src/albert/collections/attachments.py
+++ b/src/albert/collections/attachments.py
@@ -126,6 +126,9 @@ class AttachmentCollection(BaseCollection):
             ).data
         )
 
+        if "metadata" not in updated.model_fields_set:
+            return PatchPayload(data=patch_data)
+
         existing_meta = existing.metadata or AttachmentMetadata()
         updated_meta = updated.metadata or AttachmentMetadata()
 
--- a/src/albert/collections/attachments.py
+++ b/src/albert/collections/attachments.py
@@ -12,7 +12,11 @@ from albert.collections.notes import NotesCollection
 from albert.core.shared.identifiers import AttachmentId, DataColumnId, InventoryId, ProjectId
 from albert.core.shared.models.patch import PatchDatum, PatchOperation, PatchPayload
 from albert.core.shared.types import MetadataItem
-from albert.resources.attachments import Attachment, AttachmentCategory, AttachmentMetadata
+from albert.resources.attachments import (
+    Attachment,
+    AttachmentCategory,
+    AttachmentMetadata,
+)
 from albert.resources.files import FileCategory, FileNamespace
 from albert.resources.hazards import HazardStatement, HazardSymbol
 from albert.resources.notes import Note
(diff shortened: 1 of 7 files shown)


prasad-albert:
> ## Code Review
> ### Issues Found
> * **[src/albert/collections/base.py:161]** Guard condition `if attribute not in updated.model_fields_set and new_value is None` is weaker than intended. OPINIONS.md (added in this same PR) states "Gate the diff on `attribute in updated.model_fields_set`, never on `value is None` alone" — yet the condition does both, meaning an unset field whose type has a non-None default (e.g. `default=[]`) will not be skipped and will emit a spurious diff op. `tasks.py:286` correctly uses just `if attribute not in updated.model_fields_set: continue` — `base.py` should match it and drop the `and new_value is None` clause.
> 
> ### Summary
> 1 issue found. The incomplete guard in the shared `_generate_patch_payload` path contradicts the policy OPINIONS.md introduces and leaves the same class of silent-deletion bug reachable for any future field with a non-None default.

fixed this issue -- re-review the PR

prasad-albert:
> > ## Code Review
> > ### Issues Found
> > 
> > * **[src/albert/collections/base.py:161]** Guard condition `if attribute not in updated.model_fields_set and new_value is None` is weaker than intended. OPINIONS.md (added in this same PR) states "Gate the diff on `attribute in updated.model_fields_set`, never on `value is None` alone" — yet the condition does both, meaning an unset field whose type has a non-None default (e.g. `default=[]`) will not be skipped and will emit a spurious diff op. `tasks.py:286` correctly uses just `if attribute not in updated.model_fields_set: continue` — `base.py` should match it and drop the `and new_value is None` clause.
> > 
> > ### Summary
> > 1 issue found. The incomplete guard in the shared `_generate_patch_payload` path contradicts the policy OPINIONS.md introduces and leaves the same class of silent-deletion bug reachable for any future field with a non-None default.
> 
> fixed this issue -- re-review the PR

@claude re-review this PR, addressed your comments

sumedhb1995 (approved):
Awesome!! This changes update semantics for anyone who relied on "construct a fresh partial object, call update, and have omitted fields be deleted." Could you please ensure that, at-least internally, no-one was relying on this paradigm? That pattern is unusual (the documented flow is fetch-modify-save), but it is a behavior change nonetheless. I would consider at the very least including this in the some breaking changes or similar change notes to let any external consumers of the SDK know

sumedhb1995 (commented):
Just a heads up, I think `smart_datasets.py` defines its own `_generate_patch_payload` that does not delegate to base and has no `model_fields_set` guard. That will need its own separate fix

prasad-albert:
> Awesome!! This changes update semantics for anyone who relied on "construct a fresh partial object, call update, and have omitted fields be deleted." Could you please ensure that, at-least internally, no-one was relying on this paradigm? That pattern is unusual (the documented flow is fetch-modify-save), but it is a behavior change nonetheless. I would consider at the very least including this in the some breaking changes or similar change notes to let any external consumers of the SDK know

It is a behavior change -- I will be explicitly adding this in the release notes and communicate to internal users.

prasad-albert:
> Just a heads up, I think `smart_datasets.py` defines its own `_generate_patch_payload` that does not delegate to base and has no `model_fields_set` guard. That will need its own separate fix

added a fix for this in the same PR

PR #562, opened 2026-06-24
docs(agents): record unset-vs-empty rule for update payloads

prasad-albert:
Adds `OPINIONS.md` capturing the non-obvious rule for PATCH/`update()` payloads: a field the caller never set must be left untouched, gated on `model_fields_set` (never on `value is None`). Distinguishes unset (no-op) from explicit `None` (delete) and explicit `[]`/`{}` (clear).

Links it from `AGENTS.md` Coding Standards so it is read before changing update methods or patch-payload generators.

Docs-only. The corresponding code fix is in #561.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,6 +21,7 @@ Single source of truth for coding-agent guidance in this repo.
 
 ## Coding Standards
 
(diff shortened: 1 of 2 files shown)


prasad-albert:
Superseded: code fix in #561 (merged), docs in #549 (`92757b8b`) which added a fuller `OPINIONS.md` and `AGENTS.md` integration.

PR #563, opened 2026-06-24
fix(resources): tolerate None in Lot serializers and KetcherContent.data

prasad-albert:
**Lot serializers** — `serialize_initial_quantity` and `serialize_cost` called `_format_decimal()` unconditionally. Both fields are optional (`float | None`), so passing `None` crashed at serialization time. Return `None` when the value is absent; `lots.create` uses `exclude_none=True` so the field is omitted from the payload cleanly.

**KetcherContent.data** — field was typed `str` with `default=None`, causing Pydantic to reject reassignment of `None` under `validate_assignment`. Widened to `str | None` to match the actual optional semantics. Field is excluded from serialization so no wire format change.

### DIFF
--- a/src/albert/resources/lots.py
+++ b/src/albert/resources/lots.py
@@ -143,13 +143,13 @@ class Lot(BaseResource):
             formatted = formatted.rstrip("0").rstrip(".")
         return formatted
 
-    @field_serializer("initial_quantity", return_type=str)
+    @field_serializer("initial_quantity", return_type=str | None)
     def serialize_initial_quantity(self, initial_quantity: NonNegativeFloat):
-        return self._format_decimal(initial_quantity)
+        return self._format_decimal(initial_quantity) if initial_quantity is not None else None
 
-    @field_serializer("cost", return_type=str)
+    @field_serializer("cost", return_type=str | None)
     def serialize_cost(self, cost: NonNegativeFloat):
-        return self._format_decimal(cost)
+        return self._format_decimal(cost) if cost is not None else None
 
     @field_serializer("inventory_on_hand", return_type=str)
     def serialize_inventory_on_hand(self, inventory_on_hand: NonNegativeFloat):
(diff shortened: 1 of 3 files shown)


sumedhb1995 (approved):
Awesome, thank you!

src/albert/resources/lots.py
```diff
@@ -143,13 +143,13 @@ def _format_decimal(value: NonNegativeFloat) -> str:
             formatted = formatted.rstrip("0").rstrip(".")
         return formatted
 
-    @field_serializer("initial_quantity", return_type=str)
+    @field_serializer("initial_quantity", return_type=str | None)
     def serialize_initial_quantity(self, initial_quantity: NonNegativeFloat):
-        return self._format_decimal(initial_quantity)
+        return self._format_decimal(initial_quantity) if initial_quantity is not None else None
```
sumedhb1995, line 148: Hmm actually on looking at this again, I think this only half resolves the issue. The agent currently coerces None → 0.0, and there are two reasons: the serializer crashes on None and "the Lot API requires initialQuantity and cost even though the SDK marks them optional." This changes fixes the crash, but if the worker drops the 0.0 coercion, create serializes with exclude_none=True, the field is omitted entirely, and I believe the API will reject the missing field. Can you confirm? I think this might need to be 0.0

sumedhb1995 (commented):
Please confirm that API will accept `None`

src/albert/resources/lots.py
prasad-albert: The API doesn't enforce initialQuantity on POST it creates the lot without it, but GET /lots then 500s because the stored record fails response schema validation and the UI can't render. Filed [EX-2338](https://linear.app/albert-invent/issue/EX-2338/api-lot-coerce-missing-initialquantity-to-0-on-post-apiv3lots) with the backend team to coerce missing initialQuantity to 0 server-side so we don't need to manage this in the SDK.

prasad-albert: Also for non-task lot (when task_id is None) -> intial_quantity is required, whereas for task lot (with task_id set), this is optional. Until the backend addresses the linear issue -- we can let the agent coerce to 0.0 for non-task lots

sumedhb1995 (approved):

PR #565, opened 2026-06-24
fix(tasks): mirror parent_id into Project body on create

prasad-albert:
Creating a task under a project requires both a `parentId` query param and a `Project` body field. The SDK sent only the query param, so the task was created without a project association.

When `parent_id` is set and `project` is not explicitly provided, `create()` now mirrors `parent_id` into the body as `Project: {id}` before serializing. Callers who set `project` explicitly are unaffected.


prasad-albert:
Reverted — parent_id is not always a project ID so auto-mirroring is unsafe. Callers should set project explicitly. Closing without merge.

PR #567, opened 2026-06-24
fix(workflows): allow intervals on parameter groups without an id

prasad-albert:
`ParameterGroupSetpoints` raised a `ValueError` when `id` was `None` and any parameter setpoint had `intervals` set. This crashed `get_all` deserialization for any page containing such a workflow.

The API returns workflows in this state, so the check was overly strict. Removed the validator block — the API is the authority on valid workflow shapes.


prasad-albert:
Consolidated into #566 (fix/workflow-fixes).

PR #568, opened 2026-06-24
fix(workflows): handle duplicate detection and relax intervals validator

prasad-albert:
**Duplicate detection (`existingAlbertId`):** when a workflow with identical setpoints already exists, the API returns `{"existingAlbertId": "WFL…"}` with no other fields. Constructing `Workflow(**x)` crashed because `name` is required. Detect this shape (by absence of `name`) and fetch the full workflow via `get_by_id` instead.

**Intervals validator:** `ParameterGroupSetpoints` raised a `ValueError` when `id` was `None` and any parameter setpoint had `intervals` set. This crashed `get_all` deserialization for any page containing such a workflow. The API returns workflows in this state, so the check was overly strict. Removed the validator — the API is the authority on valid workflow shapes.

### DIFF
--- a/src/albert/collections/workflows.py
+++ b/src/albert/collections/workflows.py
@@ -69,7 +69,13 @@ class WorkflowCollection(BaseCollection):
                 for x in workflows
             ],
         )
-        return [Workflow(**x) for x in response.json()]
+        results = []
+        for x in response.json():
+            if "existingAlbertId" in x and len(x) == 1:
+                results.append(self.get_by_id(id=x["existingAlbertId"]))
+            else:
+                results.append(Workflow(**x))
+        return results
 
     def _hydrate_parameter_groups(self, *, workflow: Workflow) -> None:
         """Ensure parameter setpoints are fully populated for each parameter group with an id.
--- a/src/albert/collections/workflows.py
+++ b/src/albert/collections/workflows.py
@@ -71,7 +71,7 @@ class WorkflowCollection(BaseCollection):
         )
         results = []
         for x in response.json():
-            if "existingAlbertId" in x and len(x) == 1:
+            if "existingAlbertId" in x and "name" not in x:
                 results.append(self.get_by_id(id=x["existingAlbertId"]))
             else:
                 results.append(Workflow(**x))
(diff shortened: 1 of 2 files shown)


src/albert/collections/workflows.py
```diff
@@ -69,7 +69,13 @@ def create(self, *, workflows: list[Workflow]) -> list[Workflow]:
                 for x in workflows
             ],
         )
-        return [Workflow(**x) for x in response.json()]
+        results = []
```
sumedhb1995, line 72: This changes looks good!

src/albert/resources/workflows.py
```diff
@@ -250,14 +250,6 @@ def validate_identifiers(self):
         if self.parameter_group is not None and self.id is None:
             object.__setattr__(self, "id", self.parameter_group.id)
 
-        if self.id is None:
```
sumedhb1995, line 253: What motivated this change? The error that we were specifically seeing was:
```
Value error, Parameter PRM5: Special parameters cannot have 'intervals'
```
That validator was already removed on main by PR #508 (fix(workflows): allow Special parameters to have Intervals). 

It would also be good to address these for workflows but understood if this needs API support first:
**WorkflowCollection — missing search**: ``WorkflowCollection`` has no ``search``
  method. The agent can only use ``workflow_get_all`` (which returns every workflow)
  or ``workflow_get_by_id`` (which requires a known ID). A ``search(query=...)``
  method backed by the existing workflows search endpoint would let the agent resolve
  workflow names without paging the entire collection.

``WorkflowCollection`` has no ``delete`` method, so fixture-created or agent-
     created workflows cannot be cleaned up programmatically.

sumedhb1995 (approved):
`create` change looks good. Unsure about the `validate_identifiers` one as the issue we were facing already looks like it got resolved but the change here still looks valid so approving

src/albert/resources/workflows.py
prasad-albert: The API doesn't require id on a ParameterGroup -- this was an SDK-only guard with no API backing

PR #570, opened 2026-06-25
fix(property_data): type storage_key and athena on PropertyData

prasad-albert:
## Summary
- Replace `PropertyDataStorageKey | dict | None` with `StorageKeyReference` on `PropertyData.storage_key` so curve fields (`rawfile`, `s3Input`, `s3Output`) are typed and discoverable
- Remove now-unused `PropertyDataStorageKey` class
- Type `PropertyData.athena` as `CurveDBMetadata` instead of `dict[str, Any]`


lkubie (approved):
LGTM

PR #572, opened 2026-06-25
feat(session): add configurable request timeout

albert-tb:
## Problem

`AlbertSession` (the sync client's session) makes every request with no timeout, so requests inherit the `requests` default of `None` (block forever). A hung or unreachable backend can therefore block a caller indefinitely, with no way to set a session-wide deadline.

This surfaced from a react-worker investigation (Albert AI, AI-820): the worker runs sync SDK calls on a bounded thread pool, and an unbounded call permanently parks a thread. Milvus calls were bounded directly, but the SDK is the other unbounded blocking surface, and `requests` offers no session-level timeout to fix it from the outside.

## Change

Add an optional `timeout` to `AlbertSession.__init__`, stored as `self._timeout`, and apply it in `request()` via `kwargs.setdefault("timeout", self._timeout)` so an explicit per-call `timeout` still wins. Thread the parameter through `Albert.__init__` and the `from_token`, `from_sso`, and `from_client_credentials` factories, mirroring how `retries` is already plumbed.

The default stays `None`, so behavior is unchanged unless a caller opts in. A consumer can now do:

```python
Albert.from_token(base_url=..., token=..., timeout=30.0)
```

## Why this approach

`requests` has no session-level timeout: `requests.Session()` takes no `timeout` and stores none. `timeout` exists only as a per-request argument. So the idiomatic way to get a session-wide deadline is to hold a default on the session and apply it per request, which is what this does.

## Scope

Sync client only. The async client (`AsyncAlbert`) runs on `httpx`, which already applies a default timeout, so it is bounded. Adding an explicit `timeout` knob there for parity is a reasonable follow-up but is out of scope here.

## Caveat for consumers

The `requests` timeout is per attempt, and the session's retry policy multiplies it. Worst-case wall clock is roughly `(1 + retries) * timeout + backoff`. Verified empirically: a `timeout=2` against an unroutable host with the default 3 retries raised after ~9.8s, not 2s. The call is bounded and self-freeing, just not a flat `timeout`. Pair a lower retry count with the timeout if a tighter ceiling is needed.

## Testing

Repo policy is integration-style tests only (no `FakeAlbertSession` unit tests), and a timeout default is not cleanly assertable against a live API, so no new test is added. The mechanism was verified manually against an unroutable host:
- session-level `timeout` propagates to the socket (raised at the configured per-attempt deadline),
- a per-call `timeout` still overrides the session default.

`ruff format` and `ruff check` pass on the touched files.

---

### DIFF
--- a/src/albert/core/session.py
+++ b/src/albert/core/session.py
@@ -27,6 +27,12 @@ class AlbertSession(requests.Session):
         If provided, it overrides `token`.
     retries : int, optional
         The number of automatic retries on failed requests (default is 3).
+    timeout : float | tuple[float, float], optional
+        Default timeout in seconds applied to every request. Accepts a single
+        float (applied to both connect and read) or a ``(connect, read)`` tuple.
+        When ``None`` (the default), requests have no timeout and can block
+        indefinitely. A per-call ``timeout`` passed to a request always takes
+        precedence.
     """
 
     def __init__(
@@ -36,9 +42,11 @@ class AlbertSession(requests.Session):
         token: str | None = None,
         auth_manager: AlbertClientCredentials | AlbertSSOClient | None = None,
         retries: int | None = None,
+        timeout: float | tuple[float, float] | None = None,
     ):
         super().__init__()
         self.base_url = base_url
+        self._timeout = timeout
         self.headers.update(
             {
                 "Content-Type": "application/json",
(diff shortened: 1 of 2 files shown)


lkubie (approved):

src/albert/core/session.py
```diff
@@ -36,9 +40,11 @@ def __init__(
         token: str | None = None,
         auth_manager: AlbertClientCredentials | AlbertSSOClient | None = None,
         retries: int | None = None,
+        timeout: float | None = None,
```
prasad-albert: can we widen the timeout type to `float | tuple[float, float] | None` to match the `requests._types.TimeoutType` which accepts both connect/read timeouts

prasad-albert:
The failing tests are fixed in main, rebasing should make the CI green

prasad-albert (approved):
couple of comments, LGTM otherwise

PR #573, opened 2026-06-30
test(cas,substance): tighten filter and CAS ID assertions

prasad-albert:
## Summary
- `test_cas_get_all_with_filters`: after filtering by CAS number, now asserts the expected number actually appears in results (not just that the list is valid)
- `test_get_by_ids`: replaces loose `>= count` check with a membership check that verifies every requested CAS ID is present in the response


siddhant-albertinvent (approved):
approved

PR #574, opened 2026-06-30
fix(tasks): mirror parent_id to Project body on create when absent

prasad-albert:
## What

General and property task creates with `parent_id` set to a project ID now automatically include the `Project` entity link in the POST body when callers omit `project`, so tasks are linked to the project on create.

## Why

Staging verification showed that without `Project` in the request body, General and Property tasks could be created with only a query `parentId` but were not associated with the project. Batch tasks already worked without this (the API backfills `Project` from the query param).

## How

Mirror `parent_id` into `Project.id` only when:
- `Project` is not already in the serialized payload (preserves property tasks that use `parent_id=inventory_id` with an explicit `project`), and
- `parent_id` starts with `PRO` (project ID, not inventory).

## Testing

- [x] `uv run ruff format .`
- [ ] `uv run pytest tests/collections/test_tasks.py -v`
- [x] Live staging repro (one-off): batch/general/property create scenarios with and without Project mirroring

## SDK Changes

Task creates with `parent_id` set to a project ID no longer require callers to also set `project` for General and Property tasks to be linked to that project.

### DIFF
--- a/src/albert/resources/tasks.py
+++ b/src/albert/resources/tasks.py
@@ -247,7 +247,7 @@ class PropertyTask(BaseTask):
     location : SerializeAsEntityLink[Location]
         The location where the property task is performed.
     parent_id : str
-        The ID of the parent project.
+        The ID of the parent inventory item.
     blocks : list[Block]
         A list of blocks associated with the property task.
     id : str, optional
--- a/src/albert/resources/tasks.py
+++ b/src/albert/resources/tasks.py
@@ -247,7 +247,7 @@ class PropertyTask(BaseTask):
     location : SerializeAsEntityLink[Location]
         The location where the property task is performed.
     parent_id : str
-        The ID of the parent inventory item.
+        The ID of the parent project.
     blocks : list[Block]
         A list of blocks associated with the property task.
     id : str, optional
--- a/src/albert/collections/tasks.py
+++ b/src/albert/collections/tasks.py
@@ -96,7 +96,7 @@ class TaskCollection(BaseCollection):
         url = f"{self.base_path}/multi?category={task.category.value}"
         if task.parent_id is not None:
             url = f"{url}&parentId={task.parent_id}"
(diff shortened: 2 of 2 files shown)


src/albert/resources/tasks.py
```diff
@@ -247,7 +247,7 @@ class PropertyTask(BaseTask):
     location : SerializeAsEntityLink[Location]
         The location where the property task is performed.
     parent_id : str
-        The ID of the parent project.
+        The ID of the parent inventory item.
```
prasad-albert: waiting on confirmation of what exactly parentId is here? inventory or project

prakashchandrahv (approved):

src/albert/resources/tasks.py
sumedhb1995: The agent worker workaround mirrors parent_id → project for PropertyTask.This PR does two things that point the other way:

In `create()`, it mirrors parent_id → Project body only for BatchTask `(isinstance(task, BatchTask))`, not `PropertyTask`.

`BaseTask` carries both `parent_id `and `project` inherited by every task type, so both interpretations are structurally possible. The two views cannot both be correct:

If this change is right(`PropertyTask.parent_id` is an inventory item, only `BatchTask.parent_id` is a project): then the worker's `PropertyTask` mirror is actively wrong. It is stuffing an inventory id into the Project entity link. That mirror should be removed, not kept, and this is the correct fix for the BatchTask side.
If the worker is right (`PropertyTask.parent_id` is a project), then this change is incomplete (it should also mirror `PropertyTask`) and its docstring change is wrong.
I believe this needs confirmation from the task API owner or a real create test for both task types before merge

prasad-albert: confirmed with @siddhant-albertinvent -- parentId is always project ID irrespective of task type.

prasad-albert: actually, the tests are failing and they expect parent ID to be inventory ID for Property Tasks
cc: @siddhant-albertinvent

PR #564, opened 2026-06-24
fix(property-data): accept numeric values in TaskPropertyCreate.value

prasad-albert:
`TaskPropertyCreate.value` was typed `str | ImagePropertyValue | CurvePropertyValue | None`, rejecting `int` and `float` inputs at construction time.

The Albert API expects all property values as strings. Numeric inputs are now coerced to `str` via a `field_validator` before validation, so callers can pass `value=42` or `value=3.14` directly without manual conversion.

PR #578, opened 2026-07-01
fix(property-data): type PropertyDataResult.trial as Any

prasad-albert:
## Summary
- Backend PR 284 removed `trial` from the PTD result schema and it can now come back as any datatype, so the required `trial: str` field raised `ValidationError` on search responses.
- Retype `PropertyDataResult.trial` as `Any` (default `None`): accepts strings, numbers, or any type without failing, and avoids the type-specific whack-a-mole seen with `value` (#564).


siddhant-albertinvent (approved):
approved

PR #592, opened 2026-07-07
feat(attachments): add upload_and_attach_document_to_inventory_item

prasad-albert:
## Summary

- Extends `AttachmentCategory` enum with 16 new document types (CoA, TDS, RSL, EPD, PCF, BMR, MFR, NDA, Allergen, AnimalDerived, Brochure, ChangeNotification, FoodContact, Patent, ReligiousOrganicCert, SupplierQuestionnaire), each with a full UI label docstring
- Adds `upload_and_attach_document_to_inventory_item` — a generic method for attaching any document category to an inventory item, supporting the same metadata fields as the SDS method (revision date, jurisdiction, language, hazard data, storage class, etc.) all optional
- Refactors `upload_and_attach_sds_to_inventory_item` to delegate to the new generic method, eliminating duplicated upload logic

### DIFF
--- a/src/albert/collections/attachments.py
+++ b/src/albert/collections/attachments.py
@@ -420,6 +420,7 @@ class AttachmentCollection(BaseCollection):
 
         return note_collection.get_by_id(id=registered_note.id)
 
+    @validate_call
     def upload_and_attach_sds_to_inventory_item(
         self,
         *,
@@ -537,15 +538,13 @@ class AttachmentCollection(BaseCollection):
         FileNotFoundError
             If ``file_path`` does not exist.
         """
-        category_value = category.value if isinstance(category, AttachmentCategory) else category
-
         resolved_path = file_path.expanduser()
         if not resolved_path.is_file():
             raise FileNotFoundError(f"File not found at '{resolved_path}'")
 
         content_type = mimetypes.guess_type(resolved_path.name)[0] or "application/pdf"
         upload_id = self._generate_upload_id()
-        file_key = f"{inventory_id}/{category_value}/{upload_id}{resolved_path.name}"
+        file_key = f"{inventory_id}/{FileCategory.SDS.value}/{upload_id}{resolved_path.name}"
 
         file_collection = self._get_file_collection()
         with resolved_path.open("rb") as file_handle:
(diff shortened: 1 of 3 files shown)


prasad-albert:
> **Code Review**
> 
> **Issues Found**
> 
> * **[src/albert/collections/attachments.py:548]** `file_key` uses `resolved_path.name` (the full filename) instead of the extension. The original SDS code and the sibling `upload_and_attach_document_to_project` (line 631) both use `{upload_id}{extension}` where `extension = resolved_path.suffix`. As written the upload_id is concatenated with the entire filename, producing malformed keys like `INVA1/CoA/<uploadid>dontpanic.jpg`. Should be `resolved_path.suffix`.
> * **[src/albert/collections/attachments.py:557]** `sign_and_upload_file(..., category=FileCategory.SDS)` is hardcoded for every category, so a non-SDS document (e.g. `CoA`) is signed/categorized as SDS. `FileCategory` also exposes `OTHER`; non-SDS categories should not be signed as `FileCategory.SDS`.
> * **[src/albert/collections/attachments.py:423]** `@validate_call` was removed from the public method `upload_and_attach_sds_to_inventory_item`. Public collection methods use `@validate_call` per repo convention (validation still happens in the delegate, but the decorator should remain on the public entry point).
> 
> **Summary** 3 issues found. The `file_key` construction (line 548) is a behavior regression that produces incorrectly-formed storage keys, and the hardcoded `FileCategory.SDS` mislabels non-SDS documents — both should be fixed before merge.

@claude 1 and 2 are non-issue. The API is structured in such a way that currently it formats the file key with both uploadID+name, and sets file cateogry as SDS always.
Fix the 3rd issue by adding a validate_call

prasad-albert:
> ## Code Review
> ### Issues Found
> * **src/albert/collections/attachments.py:547** The storage key is built as `f"{inventory_id}/{FileCategory.SDS.value}/{upload_id}{resolved_path.name}"`, concatenating the UUID directly to the _full_ filename (e.g. `INVA123/SDS/<uuid>dontpanic.jpg`). The prior code and the sibling `upload_and_attach_document_to_project` (line 629-630) use `resolved_path.suffix` -> `<uuid>.jpg`. This should be `{upload_id}{resolved_path.suffix}`. Since `upload_and_attach_sds_to_inventory_item` now delegates here, this is also a behavior regression for SDS keys.
> 
> ### Summary
> 1 issue found. Malformed storage keys for all inventory document uploads (including SDS), a regression from the previous `.suffix`-based key format.

@claude non issue, API behaves differently.

prakashchandrahv (approved):

PR #624, opened 2026-07-17
docs: migrate resource models to attribute docstrings

sumedhb1995:
## What

Field-level documentation on ~246 SDK resource models now lives in attribute docstrings instead of class-level `Attributes` sections, so JSON schemas expose descriptions for downstream consumers (Ask Albert tool factory, `expand_schema`).

## Why

The react worker builds agent tools from SDK method signatures and Pydantic schemas. Descriptions were not reaching the agent unless duplicated manually in the factory. This migration wires SDK docs through automatically and adds guardrails so the pattern cannot regress.

## How

- Ran a codemod (`scripts/migrate_attributes_docstrings.py`) across `src/albert/resources/` and shared models
- Fixed the `HazardStatement` straggler in `hazards.py`
- Added unit tests asserting no leftover `Attributes` sections and that attribute docstrings reach `model_fields`
- Updated `AGENTS.md` with the attribute-docstring convention

## Testing

```bash
cd albert-python
uv run pytest tests/unit/test_model_attribute_docstrings.py tests/unit/test_resource_model_docstring_coverage.py -q
```

Example: `Project.model_json_schema()` now documents all 15 fields (was 1/15 before).


lkubie (approved):
LGTM

PR #635, opened 2026-07-21
feat: add has_more to AlbertPaginator + Sheet.add_parameter_group_row (AI-1086, AI-1087, AI-1071)

sumedhb1995:
## What

Albert SDK list iterators now expose a definitive `has_more` (and optional `total`) so callers know when a paginated sweep stopped early versus returning the full match set.

Also adds `Sheet.add_parameter_group_row` so Process Design parameter groups can be added via the same design-engine path the UI uses ([AI-1071](https://linear.app/albert-invent/issue/AI-1071)).

## Behavior change: project search pagination is now complete

`projects.search()` and `projects.get_all()` previously returned at most ~25 results in most cases: the backend ignored `limit` on the GET search path and returned empty pages for `offset>0`, so iteration stopped after one page. Project search now uses the same request path as the Albert UI, and pagination continues until the full match set is exhausted.

**What existing callers will notice:**

- Unbounded loops over `projects.search()` / `projects.get_all()` now iterate the full result set (potentially thousands of projects) instead of silently stopping at ~25. `projects.get_all()` additionally performs one hydration request per result: pass `max_items` if you only need a sample.
- Integrations that distinguish GET vs POST on `/api/v3/projects/search` (gateways, audit rules) will see POST for all project searches.
- **New:** after iteration, every list/search paginator exposes `has_more` (`True` when the iteration was cut off by `max_items` and more matches are known to exist) and `total` (server-reported match count, where the endpoint provides one). `get_all` hydration wrappers preserve both.

## Why

Fixes incomplete-result bugs tracked in [AI-1086](https://linear.app/albert-invent/issue/AI-1086) and [AI-1087](https://linear.app/albert-invent/issue/AI-1087). Agents were treating capped or under-filled pages as complete sets — e.g. filtering a globally capped search by location, or stopping after the first page of task properties.

Also unblocks [AI-1071](https://linear.app/albert-invent/issue/AI-1071): Process Design PRG rows must POST to `/api/v3/designs/{id}/rows` (worksheet design rows reject `type=PRG`).

## How

- Extended `AlbertPaginator` to track continuation state across offset/lastKey pages and when backends under-return a page.
- Switched project search to POST (Albert UI path) so offset/limit pagination works end-to-end.
- Updated collection modules (`tasks`, `projects`, `cas`, chat collections, etc.) to use the new semantics.
- Added `tests/core/test_pagination.py` covering under-filled pages, broken offset backends, KEY-mode cursors, and natural exhaustion.
- Added `Sheet.add_parameter_group_row(parameter_group_id=...)` posting to `/api/v3/designs/{process_design.id}/rows`.

## Testing

```bash
cd albert-python && uv run pytest tests/core/test_pagination.py -q
cd albert-python && uv run pytest tests/resources/test_sheets.py::test_add_parameter_group_row_requires_process_design -q
```


AGENTS.md
```diff
@@ -78,6 +78,10 @@ Many list/search methods use `AlbertPaginator` (`src/albert/core/pagination.py`)
 
 Expose a `max_items` parameter on public list/search methods where appropriate to allow early stopping.
 
+After iteration, `AlbertPaginator.has_more` is True when `max_items` stopped the iterator and more items are known to exist (unyielded items on the current page, or a continuation key/offset), or when response `total` exceeds items yielded. It is False after a natural end of results. Prefer this flag over comparing yielded count to `max_items` (backends can under-return a page).
+
+Offset search endpoints (projects, tasks, lots, inventories, datatemplates, parametergroups, propertydata) return a string/int `total` field — `_record_total` reads that key. Never wrap a returned paginator in `yield from` / `async for … yield` inside `get_all`; return the paginator (or `MappedPaginator` / `MetadataPreservingIterator`) so callers keep `has_more` / `total`.
```
prasad-albert, line 83: It might be helpful to add a brief decision guide here on when to use `AlbertPaginator` directly vs `MappedPaginator` vs `MetadataPreservingIterator`. The "don't wrap in `yield from`" note is already great — a small when-to-use section could make this easier for the next person touching pagination.

tests/resources/test_sheets.py
```diff
@@ -326,6 +329,70 @@ def test_add_and_remove_blank_rows(seeded_sheet: Sheet):
...
+def test_add_parameter_group_row(
+    seeded_sheet: Sheet,
+    seeded_parameter_groups: list,
+):
+    """Test adding a parameter group row to Process Design when the sheet has one."""
+    if seeded_sheet.process_design is None:
+        pytest.skip("Seeded sheet has no Process Design section")
+
+    pg = seeded_parameter_groups[0]
+    row = seeded_sheet.add_parameter_group_row(parameter_group_id=pg.id)
```
prasad-albert: Looks like CI is failing here because the default `reference_id="ROW1"` may not be valid for Process Design on the seeded sheet (`invalid existing reference row order`). I wonder if the test could pass a row from `seeded_sheet.process_design.rows[0]` instead — and possibly default to the first Process Design row in the SDK if that would be safer for callers too.

src/albert/collections/projects.py
```diff
@@ -503,39 +491,45 @@ def get_all(
         Returns
         -------
         Iterator[Project]
-            An iterator of fully populated Project entities.
+            An iterator of fully populated Project entities. Preserves ``has_more`` /
+            ``total`` from the underlying search paginator.
```
prasad-albert, line 504: It might be worth adding a short usage example here (oron `search`) showing how to keep the paginator reference and check `has_more` / `total` after iteration. Even a brief docs note could help callers discover this.

src/albert/collections/parameter_groups.py
```diff
@@ -296,22 +296,30 @@ def get_all(
...
+
+        def _hydrate(item: ParameterGroupSearchItem) -> ParameterGroup | None:
             try:
-                yield self.get_by_id(id=item.id)
+                return self.get_by_id(id=item.id)
             except AlbertHTTPError as e:  # pragma: no cover
                 logger.warning(f"Error fetching parameter group {item.id}: {e}")
+                return None
+
+        return MappedPaginator(
```
prasad-albert, line 313: I noticed `users` and `storage_locations` still hydrate via `deserialize` inside `AlbertPaginator`, which should preserve metadata correctly but differs from the `MappedPaginator` pattern used here. No rush, but it might be nice to align those in this PR or a small follow-up if you think it's worth the consistency.

prasad-albert (approved):
LGTM — nice work on the pagination completeness semantics and the project search fix. A few small suggestions below; none are blockers from my side.

prasad-albert (approved):

AGENTS.md
sumedhb1995: Good call — added a short decision table under Pagination in `AGENTS.md` covering when to return `AlbertPaginator` vs `MappedPaginator` vs `MetadataPreservingIterator` (including that hydrate-in-`deserialize` is fine when the list endpoint is already the source).

tests/resources/test_sheets.py
sumedhb1995: Fixed both sides: the test now anchors on `seeded_sheet.process_design.rows[0].row_id`, and `add_parameter_group_row` defaults `reference_id` to the first Process Design row instead of hardcoding `"ROW1"`.

src/albert/collections/projects.py
sumedhb1995: Added examples on both `search` and `get_all` that keep the paginator reference and check `has_more` / `total` after iteration.

src/albert/collections/parameter_groups.py
sumedhb1995: Agreed it's worth consistency eventually, but leaving as-is for this PR — hydrate-in-`deserialize` already preserves `has_more` / `total`, and those collections don't have a separate search→hydrate split. Captured the distinction in the new AGENTS decision guide; happy to align users/storage_locations in a small follow-up if useful.

sumedhb1995: Went ahead and converged in this PR — `users.get_all` and `storage_locations.get_all` now hydrate via `MappedPaginator` like the other collections, and the AGENTS decision guide reflects that.

PR #650, opened 2026-07-28
fix(sdk): tolerate malformed entity-link metadata in parameter groups

albert-tb:
## What
Parameter group search, get_by_id, get_by_ids, and get_all now succeed for tenants with parameter groups whose metadata contains malformed entity-link values (e.g. `[{}]`). Previously any one such row silently zeroed out the entire result page.

## Why
FEP-975: Ask Albert's parameter_group_search and parameter_group_get_all tools were returning 0 results despite thousands existing on TEN1/dev. Root cause: `MetadataItem` can't parse `[{}]` (an entity-link list entry missing the required `id`), and AlbertPaginator deserialized a full 1000-item page in one `list(...)` call, so the ValidationError from one bad row discarded all 999 good ones.

## How
- Added a `mode="before"` field validator on `ParameterGroup.metadata` and `ParameterGroupSearchItem.metadata` that drops entity-link entries missing `id`, following the existing IN-10 precedent in this file (`ParameterValue.value`).
- `AlbertPaginator._create_iterator` now retries per-item only when the batched deserialize raises `ValidationError`, preserving the single-call happy path (important for collections like `workflows`/`inventory` whose `deserialize` calls `get_by_ids` as one batched HTTP request). Falls back to per-item deserialization and logs+skips the specific bad row.
- Scoped to `ValidationError` only (not a bare `except Exception`) so transport failures from I/O-performing `deserialize` callables still propagate instead of being silently swallowed.
- Verified against real TEN1/dev data: this also surfaced a second, unmodeled metadata shape (nested `[[{...}]]` entity-link lists) on 3 other parameter groups. The paginator fallback now tolerates it (skips + warns) but the type-level fix doesn't model it -- noted as a known remaining case, not a full fix.

## Testing
- `tests/collections/test_parameter_groups.py`: two new unit tests constructing both models directly from the malformed raw payload captured off dev (PRG23111).
- `tests/core/test_pagination.py` (new file): paginator test with a fake session/page where one item fails pydantic validation, asserting the other items still yield and a warning is logged.
- Manually verified end-to-end against TEN1/dev: `search(max_items=1500)` now returns all 1500 items including the previously-failing row, and `get_by_id` on it succeeds.


albert-tb:
@claude please review this pr

prasad-albert (approved):

PR #658, opened 2026-08-03
fix(inventory): preserve metadata list item names

albert-tb:
## What

Inventory metadata list values now preserve the names returned by the Albert API when inventory items are retrieved and serialized.

## Why

The API already returns readable names for list-based custom-field values, but the SDK's `EntityLink` serialization excluded those names. This caused consumers such as the inventory data agent to receive only list-item IDs. This addresses AI-1336.

## How

- Use an inventory-specific metadata type that accepts named links while remaining compatible with existing ID-only links.
- Do not add extra list or custom-field API calls.
- Keep the existing `InventoryItem.metadata` structure unchanged for callers.

## Testing

- `uv --directory albert-python run ruff format src/albert/resources/inventory.py tests/collections/test_inventory.py`
- `uv --directory albert-python run ruff check src/albert/resources/inventory.py tests/collections/test_inventory.py`
- Live integration test with `-n 4`: passed.

### DIFF
--- a/src/albert/resources/inventory.py
+++ b/src/albert/resources/inventory.py
@@ -8,8 +8,8 @@ from pydantic import Field, field_validator, model_validator
 from albert.core.base import BaseAlbertModel
 from albert.core.shared.enums import SecurityClass
 from albert.core.shared.identifiers import InventoryId
-from albert.core.shared.models.base import AuditFields, EntityLink, EntityLinkWithName
-from albert.core.shared.types import SerializeAsEntityLink
+from albert.core.shared.models.base import AuditFields
+from albert.core.shared.types import MetadataItem, SerializeAsEntityLink
 from albert.resources._mixins import HydrationMixin
 from albert.resources.acls import ACL
 from albert.resources.cas import Cas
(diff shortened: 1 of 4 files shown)


src/albert/resources/inventory.py
```diff
@@ -19,6 +19,10 @@
 from albert.resources.tagged_base import BaseTaggedResource
 from albert.resources.tags import Tag
 
+InventoryMetadataItem = (
```
prasad-albert: Do you think we can fix it at the shared MetadataItem level (EntityLinkWithName in types.py) instead of an inventory-specific alias (as long as we're not breaking existing functionality)? That would keep things consistent across collections and avoid duplicating a type.

From what I can tell, writes should be fine - PATCH already sends bare IDs, and create paths use exclude_none/exclude_unset, so id-only links still dump as {"id": "..."}. The change would mainly affect read/export, where preserving names seems desirable everywhere.

albert-tb: good call!

prasad-albert (approved):

PR #663, opened 2026-08-07
fix(substance_v4): stop search pagination at final page

prasad-albert:
## What

Uncapped `client.substances_v4.search()` now completes without error and exposes `paginator.total`. `get_or_create` for locations and storage locations is idempotent under parallel test runs and search-index lag.

## Why

Substance v4 search ignored nested `pagination` metadata and over-requested past the final page (400). Location/storage-location `get_or_create` could miss existing records and attempt a duplicate create, failing with "Name already exists" or sending read-only audit fields on storage location POST.

## How

- Substance v4 paginator reads `pagination.total` and stops when `pagination.lastKey` is omitted.
- Location `get_or_create` returns by id when set, uses exact name lookup with fuzzy fallback, and retries lookup after duplicate-name 400s.
- Storage location `create` excludes read-only fields; `get_or_create` falls back to scanning the parent location when name filters lag.

## Testing

- [x] `uv run ruff format .`
- [x] `uv run pytest tests/core/test_pagination.py -k substance_v4 -v`
- [x] `uv run pytest tests/collections/test_substance_v4.py::test_search_pagination -v -n 4`
- [x] `uv run pytest tests/collections/test_locations.py tests/collections/test_storage_locations.py -v -n 4`

## SDK Changes

Fixed substance v4 search pagination and hardened location/storage-location `get_or_create` against duplicate-create failures during parallel integration tests.

### DIFF
--- a/src/albert/collections/locations.py
+++ b/src/albert/collections/locations.py
@@ -5,6 +5,7 @@ from albert.core.pagination import AlbertPaginator
 from albert.core.session import AlbertSession
 from albert.core.shared.enums import PaginationMode
 from albert.core.utils import ensure_list
+from albert.exceptions import BadRequestError
 from albert.resources.locations import Location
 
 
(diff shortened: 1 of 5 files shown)


prasad-albert:
This PR also fixes idempotency on get_or_create for locations, storage_locations

albert-tb (approved):
looks good! thank you for fixing this!!

PR #670, opened 2026-08-11
fix(hazards): support model-selected symbol status

albert-tb:
## What

Attachments containing hazard symbols selected by a model now deserialize successfully through the SDK.

## Why

The released SDK rejects the platform's `From Model` hazard-symbol status, causing `attachments.get_by_id()` to fail for attachments such as ATT111191. This fixes [SDK-86](https://linear.app/albert-invent/issue/SDK-86/hazardsymbolstatus-enum-missing-from-model-attachmentsget-by-id-raises).

## How

Adds the `From Model` wire value to `HazardSymbolStatus` and covers it in the attachment regression path. The previously proposed SDK-64 filter was removed after stable 1.42.0 testing showed the backend already omits the relevant unknown CAS record; valid-format unknown CAS 422s belong to REG-1129/API handling.

## Testing

- [x] Targeted Ruff format checks for changed files
- [x] Targeted Ruff lint checks for changed files
- [x] Local source regression: `HazardSymbol(status="From Model")` parses
- [x] Local ATT111191-style attachment metadata regression parses
- [x] Stable `albert==1.42.0` comparison reproduced the pre-fix validation failure
- [ ] `uv run pytest tests/collections/test_attachments.py -v` (requires client-credentials test environment)

## SDK Changes

The SDK now accepts model-selected hazard-symbol metadata returned by the platform.

### DIFF
--- a/src/albert/collections/substance_v4.py
+++ b/src/albert/collections/substance_v4.py
@@ -186,15 +186,7 @@ class SubstanceV4Collection(BaseCollection):
             params["classificationType"] = classification_type
 
         response = self.session.get(self.base_path, params=params)
-        result = SubstanceV4Response.model_validate(response.json())
-        if cas_ids:
-            requested_cas_ids = set(cas_ids)
-            result.substances = [
-                substance
-                for substance in result.substances
-                if substance.is_cas or substance.cas_id not in requested_cas_ids
-            ]
-        return result
+        return SubstanceV4Response.model_validate(response.json())
 
     @validate_call
     def get_by_id(
--- a/src/albert/collections/substance_v4.py
+++ b/src/albert/collections/substance_v4.py
@@ -186,7 +186,15 @@ class SubstanceV4Collection(BaseCollection):
             params["classificationType"] = classification_type
 
         response = self.session.get(self.base_path, params=params)
-        return SubstanceV4Response.model_validate(response.json())
+        result = SubstanceV4Response.model_validate(response.json())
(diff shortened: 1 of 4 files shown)


src/albert/collections/substance_v4.py
```diff
@@ -186,7 +186,15 @@ def get_by_ids(
             params["classificationType"] = classification_type
 
         response = self.session.get(self.base_path, params=params)
-        return SubstanceV4Response.model_validate(response.json())
+        result = SubstanceV4Response.model_validate(response.json())
```
prasad-albert: The substance placeholder filter appears redundant — `get_by_ids(cas_ids=["DUMMY-CAS-NO", ...])` already returns zero fabricated records on dev and prod. The backend omits unknown CAS IDs rather than returning isCas=false stubs for bare casIDs lookups. The filter may also incorrectly drop intentional trade-secret results (isCas=false) when non-CAS identifiers are passed via cas_ids. Recommend dropping the substance_v4 change and merging only the HazardSymbolStatus.FROM_MODEL fix (SDK-86), which is a real validation gap.

albert-tb:
Removed the SDK-64 filter after testing stable albert==1.42.0 against production: DUMMY-CAS-NO already returns zero v4 records, while valid-format unknown CAS 422s are the separate REG-1129/API issue. This PR now contains only the SDK-86 fix.

prasad-albert (approved):

PR #686, opened 2026-08-17
feat(collections): add missing search filters for tasks and lots

prasad-albert:
## What

Task search gains UI-oriented filters (facets, witness status, due date, attachments, and more). Lot search gains a POST path for metadata/custom-field and facet filters.

## Why

Task and lot APIs document filters the SDK did not expose; lots POST `/search` was entirely unavailable.

## How

- **Tasks**: 13 new filters on `search()`; `get_all()` inherits them via delegation.
- **Lots**: POST `/search` when `metadata_filters` or `custom_fields` is set, with POST-only filters; GET path unchanged.

## Testing

- [x] `uv run ruff format .`
- [x] `uv run pytest tests/collections/test_tasks.py tests/collections/test_lots.py -n 4 -v`

## SDK Changes

Expands `TaskCollection.search()` and adds POST-based lot search filters on `LotCollection.search()`.

### DIFF
--- a/src/albert/collections/lots.py
+++ b/src/albert/collections/lots.py
@@ -1,6 +1,6 @@
 from collections.abc import Iterator
 from decimal import Decimal
-from typing import Literal
+from typing import Any, Literal
 
 from pydantic import validate_call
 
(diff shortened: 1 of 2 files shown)


prasad-albert:
> ## Code Review
> ### Issues Found
> * **[src/albert/collections/tasks.py:832-841]** The new facet/contains filters (`containsField`, `containsText`, `facetField`, `facetText`, `witnessStatus`, `hasAttachment`, `dueDateDuration`, `collaboratorPopUp`, `linkedTo`, `additionalField`, `sourceField`) are placed in the plain `params` dict, so they are sent as **GET query params** whenever `metadata_filters` is not provided. In `lots.py` the same filter families (`containsField`, `containsText`, `facetField`, `facetText`, etc.) were deliberately routed through `post_only_params` and lot search now always POSTs. This is a clear asymmetry: if the tasks `/search` backend only honors these filters via POST (as the lots change implies), a caller doing `search(facet_field=..., contains_field=...)` without `metadata_filters` will silently have them ignored on the GET path. Please confirm the tasks GET endpoint actually accepts these params; otherwise route them through POST like lots.
> 
> ### Summary
> 1 issue found. Not a hard runtime error, but a likely correctness/consistency gap where task facet/contains filters may be silently dropped on the GET path. The rest of the change (POST payload construction, `@validate_call`, wire-key casing, `None`-filtering via `_serialize_payload`, `get_all` delegation, docstrings, and conventional-commit title) looks correct.

Confirmed against api-task OpenAPI: GET /tasks/search accepts these query params. Lots routed them POST-only because lot GET doesn’t list them. No change needed on the task GET path.

prakashchandrahv (approved):

PR #689, opened 2026-08-17
feat(sds): add generate_sds for formula inventory items

prasad-albert:
## What

Callers can generate a GHS Safety Data Sheet (JSON plus a short-lived PDF URL) for a **formula** inventory item via `client.sds.generate_sds`, with tenant-specific lookups for jurisdiction, language, product type, physical state, and legal entity.

## Why

SDS generate was only available in the Albert UI. SDK callers had no way to produce an SDS for a formula without wrapping internal document-generator and rule-engine services themselves.

## How

`generate_sds` matches the UI path: it loads the inventory name, unpacks the formula for `substances` / `casLevelSubstances` / `inventorySDSList`, and POSTs to document-generator. Callers supply lookup **values** (not display names) on `SDSRequest`; they cannot pass name or composition lists. Non-formula inventory items raise `AlbertException`. Unpack models now reject unknown keys so the wire shape stays explicit. Do not use attachments jurisdiction/language codes for generate.

## Testing

- [x] `uv run ruff format .`
- [x] `uv run pytest tests/resources/test_sds.py -v`
- [x] `uv run pytest tests/collections/test_product_design.py -v`

Lookups were exercised live. Full generate against a seeded SDK-Test formula returns the same `ERROR_MISSING_INFO_CAS` as the UI when CAS hazard data is missing. `test_generate_sds_unpacks_formula` is marked `slow`.

## SDK Changes

- New beta collection: `client.sds` with `generate_sds`, jurisdiction/language/product/physical-state/legal-entity lookups, and `get_field_options`.
- `SDSRequest` / `GeneratedSDS` resources. Formula-only; `albert_id` accepts `InventoryId` (auto-prefixes `INV`).
- Unpack (`UnpackedProductDesign`) now includes `substances` and related CAS fields, and forbids extra keys.


albert-tb (approved):
looks good!

Issue #664, opened 2026-08-10
property_data.update_or_create_task_properties does not update calculated columns


walter1-henkel:
when using the property_data.update_or_create_task_properties() method to add values to a task that also contains calculated columns, these columns stay blank instead of being filled automatically.

when using add_properties_to_task() it works as expected.

attached is a code snipped, however it is very similar to the example code provided in the documentation.

[Issue20260810_minimun_example.py](https://github.com/user-attachments/files/30897069/Issue20260810_minimun_example.py)

prasad-albert:
@walter1-henkel This should be fixed in v1.47.1

PR #693, opened 2026-08-19
fix(property_data): recalculate columns on task property upsert

prasad-albert:
## What

`update_or_create_task_properties` now fills calculated columns after writing values, including when updating an existing trial. Passing `trial_number` on an empty placeholder row creates a real trial instead of failing.

## Why

Callers writing results into a data template with formulas saw calculated cells stay blank or stale (`#664`). `add_properties_to_task` already filled those columns; upsert did not.

## How

Write input values first, re-read the trial, then patch formula columns from the fresh values. New rows go through `add_properties_to_task` so image and curve creates get the same calculation path. Placeholder `trial_number` values (no stored data yet) are omitted so the first row can be created.

## Testing

- [x] `uv run ruff format .`
- [ ] `uv run pytest tests/collections/test_property_data.py -v`
- Live repro on `dev.albertinventdev.com`: first write `5,16` → calc `9`; update existing trial to `10,16` → calc `14`; skeleton `trial_number=1` no longer 400s.
- Added `test_update_or_create_task_property_calculation_evaluation`. Full-file pytest setup failed on an unrelated `POST /parametergroups` seeding error (`One or more Units do not exist!`).

## SDK Changes

`update_or_create_task_properties` now recalculates formula columns after upserts, matching `add_properties_to_task`.

### DIFF
--- a/src/albert/collections/property_data.py
+++ b/src/albert/collections/property_data.py
@@ -1056,6 +1056,44 @@ class PropertyDataCollection(BaseCollection):
             get_task_block_properties=self.get_task_block_properties,
         )
 
+    def _apply_calculated_task_property_patches(
+        self,
+        *,
+        inventory_id: InventoryId,
+        task_id: TaskId,
+        block_id: BlockId,
+        lot_id: LotId | None,
+        properties: list[TaskPropertyCreate],
+        return_scope: ReturnScope,
+    ) -> list[TaskPropertyData]:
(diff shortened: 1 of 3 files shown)


prasad-albert:
fixes https://github.com/albert-labs/albert-python/issues/664

lkubie (approved):
I'm having bad flashbacks of originally writing this. Much cleaner!

PR #694, opened 2026-08-19
feat(workflows): add search method

prasad-albert:
## What

Callers can search workflows by text, IDs, status, data templates, parameter groups, parameters, units, and bound `parameter_set` constraints. Hits are lightweight [`WorkflowSearchItem`](https://github.com/albert-labs/albert-python) objects; call `hydrate()` (or `get_by_id`) for the full setpoint payload.

## Why

Workflows previously only supported list-all and get-by-id. Looking up a workflow by name meant paging the entire collection.

## How

- `search()` returns `WorkflowSearchItem`, not `Workflow`, because search hits omit setpoints and intervals.
- The ID filter is `ids=` (maps to wire `albertId`).
- Offset pagination uses page size 100, the API maximum. `offset`/`limit` are not public.
- `get_all` is unchanged (Dynamo KEY listing). Facets are deferred to a follow-up.

## Testing

- [x] `uv run ruff format .`
- [x] `uv run pytest tests/collections/test_workflows.py -v`

Search tests are marked `xfail(raises=NotFoundError, strict=False)` until `POST /api/v3/workflows/search` is deployed ([api-workflow#129](https://github.com/MoleculeEngineering/api-workflow/pull/129)). They will XPASS once the route is live; remove the decorator then. Existing get_all / get_by_id / create tests pass.

## SDK Changes

Added `client.workflows.search(...)` for full-text and filtered workflow lookup, returning `WorkflowSearchItem` with `hydrate()` to the full `Workflow`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -125,7 +125,7 @@ Offset search endpoints (projects, tasks, lots, inventories, datatemplates, para
 
 Rule of thumb: if you already have an `AlbertPaginator`, return it. Reach for `MappedPaginator` / `MetadataPreservingIterator` only when a plain generator would strip `has_more` / `total`.
 
-**`offset` and `limit` are never exposed as method parameters.** They are internal pagination state managed entirely by `AlbertPaginator`. Never add `offset` or `limit` to a public method signature or docstring. Use `max_items` as the only caller-facing pagination control.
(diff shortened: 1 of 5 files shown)


prakashchandrahv (approved):

prasad-albert:
> ## Code Review
> ### Issues Found
> * **tests/collections/test_workflows.py:32-105** The 5 new `test_workflow_search_*` tests have no `xfail` markers, but the PR description states they should be marked `xfail(raises=NotFoundError, strict=False)` until `POST /api/v3/workflows/search` is deployed. As written, these integration tests will hard-fail with `NotFoundError` in CI until the route is live. Either add the markers as described, or confirm the endpoint is deployed.
> 
> ### Summary
> 1 issue found. The SDK implementation (`search()` plus the `WorkflowSearchItem`/`WorkflowParameterSet` models) is correct and matches the existing `parameter_groups.search` pattern; the only concern is the test suite diverging from the PR description, which will break CI if the endpoint is not yet available.

@claude the API is deployed and live so we're good

PR #706, opened 2026-08-24
feat(data_templates): add get_document_version_history method

prasad-albert:
## What

Callers can now retrieve the full version history for any document attached to a data template via `client.data_templates.get_document_version_history(document_id=doc.id)`.

## Why

Ask Albert was unable to read data template document history (AI-1322). The `GET /api/v3/documents?documentId=<id>` endpoint exists in the document service but had no SDK wrapper.

## How

- Added `DocumentId` annotated type (`DOC` prefix) to `identifiers.py`, following the existing pattern for all Albert ID types.
- Added `src/albert/resources/documents.py` with `DocumentVersion`, `DocumentState`, and `DocumentClass` models mapping the full `api-document` OpenAPI contract.
- Added `get_document_version_history(*, document_id: DocumentId) -> list[DocumentVersion]` to `DataTemplateCollection`. The docstring explicitly guides callers to obtain the document ID from `DataTemplate.documents[i].id`.
- Added `docs/resources/documents.md` and the corresponding `mkdocs.yml` entry.

## Testing

- [x] `uv run ruff format .`
- [ ] `uv run pytest tests/test_data_templates.py -v`

## SDK Changes

**New:** `client.data_templates.get_document_version_history(document_id="DOC...")` returns a `list[DocumentVersion]` with fields `id`, `name`, `key`, `version_id`, `doc_version`, `document_class`, `state`, `signed_url`, `published_at`, `created`, and `updated`.

### DIFF
--- a/docs/resources/documents.md
+++ b/docs/resources/documents.md
@@ -1,0 +1,1 @@
+::: albert.resources.documents
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -216,6 +216,7 @@ nav:
           - Data Columns: resources/data_columns.md
           - Data Templates: resources/data_templates.md
           - Design: resources/design.md
+          - Documents: resources/documents.md
           - Entity Types: resources/entity_types.md
           - Enums: resources/enums.md
           - Files: resources/files.md
--- a/src/albert/collections/data_templates.py
+++ b/src/albert/collections/data_templates.py
@@ -26,7 +26,7 @@ from albert.resources.data_templates import (
     ImageExample,
     ParameterValue,
 )
-from albert.resources.documents import DocumentVersion
+from albert.resources.documents import DocumentState, DocumentVersion
 from albert.resources.parameter_groups import DataType, EnumValidationValue, ValueValidation
 from albert.utils._patch import (
     build_acl_patch_payload,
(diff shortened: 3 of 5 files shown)


sumedhb1995 (approved):
The pagination comment here looks worthwhile to address, specifically with the has_more technique applied to other collections. Otherwise looks great!

prasad-albert:
> The pagination comment here looks worthwhile to address, specifically with the has_more technique applied to other collections. Otherwise looks great!

Verified with the backend implementation -- the endpoint doesn't support pagination, caps max result to 50

prasad-albert:
> ## Code Review
> ### Issues Found
> * **[src/albert/collections/data_templates.py:1072-1073]** `get_document_version_history` reads only `response.json().get("Items", [])` with no pagination handling. The `Items` response shape is the offset-paginated endpoint shape, so a document with a long version history would be silently truncated at the backend default page size, with no indication more exist. If histories can realistically exceed one page, page through results; otherwise confirm the endpoint is non-paginated.
> 
> ### Summary
> 1 issue found. Low severity in practice (version histories are typically small), but it is the silent-truncation pattern the SDK explicitly guards against. No hard bugs: `@validate_call`, modern type hints, enum-param serialization (session converts `Enum` -> `.value`), docstrings, and the conventional-commit title all check out.

@claude the API does not support pagination

PR #721, opened 2026-09-04
feat(notebooks): add search for notebook block content

prasad-albert:
## What

`NotebookCollection` now has a `search(...)` method wrapping `GET /api/v3/notebooks/search`, so callers (Ask Albert, scripts) can find notebook **block-level** content via free text with an optional project scope, without dropping to raw `session.get`. Each hit is a new `NotebookSearchItem` model representing one matching block, not a full notebook.

## Why

Unblocks ELN content discovery across projects for agent search ([SDK-103](https://linear.app/albert-invent/issue/SDK-103/add-notebook-collection-search-support)).

## How

- `search()` builds a camelCase `params` dict (`text`, `projectId`, `sortBy`, `order`) and returns an OFFSET-mode `AlbertPaginator`; pagination is caller-controlled via `max_items` only (no public `offset`/`limit`).
- `NotebookSearchItem` maps the OpenSearch block-hit fields (`blockId`, `notebookId`, `projectId`, `blockType`, `content`, `name`, `status`, `Created`, `Updated`) with attribute docstrings; undeclared fields are ignored via the `BaseAlbertModel` default.
- No hydration on hits; callers use `get_by_id` / `get_block_by_id` for full content. `list_by_parent_id` docstring cross-links to `search` for block-content queries.

## Deferred

`globalSearch` (cross-module project search for tasks, inventory, and users on the same endpoint) is **not** included in this PR. The OpenAPI spec documents the query param but not the alternate grouped response shape (`{category, total, data}`), so we are waiting on confirmation from the API team before modeling and exposing it in the SDK.

## Testing

Added `test_search` to `tests/collections/test_notebooks.py` (existing `projects` xdist group): searches seeded notebooks by `seed_prefix` scoped to the seed project, filters hits to the fixture's notebook ids, wraps in `poll_until` for index lag, and asserts a hit with the expected `notebook_id` and non-null `block_id`.

```bash
uv run ruff format .
uv run ruff check . --fix
uv run pytest tests/collections/test_notebooks.py -v -n 4
```

Lint/format pass. Live integration run requires `ALBERT_CLIENT_ID_SDK` / `ALBERT_CLIENT_SECRET_SDK` / `ALBERT_BASE_URL`, which were not available in the build environment; the test file collects cleanly and `NotebookSearchItem` deserialization was validated against the documented wire shape.

Cake session: https://agents.ai.albertinventdev.com/sessions/ea7ea125-16b1-48f9-9f7d-5de76788e6a0

### DIFF
--- a/src/albert/collections/notebooks.py
+++ b/src/albert/collections/notebooks.py
@@ -12,7 +12,13 @@ from albert.core.base import BaseAlbertModel
 from albert.core.pagination import AlbertPaginator
 from albert.core.session import AlbertSession
 from albert.core.shared.enums import OrderBy, PaginationMode
-from albert.core.shared.identifiers import NotebookId, ProjectId, SynthesisId, TaskId
+from albert.core.shared.identifiers import (
+    NotebookId,
+    ProjectId,
+    SearchProjectId,
+    SynthesisId,
+    TaskId,
+)
 from albert.exceptions import AlbertException
 from albert.resources.files import FileNamespace
 from albert.resources.notebooks import (
@@ -172,7 +178,7 @@ class NotebookCollection(BaseCollection):
         self,
         *,
         text: str | None = None,
-        project_id: ProjectId | None = None,
+        project_id: SearchProjectId | None = None,
         sort_by: str | None = None,
         order: OrderBy | None = None,
         max_items: int | None = None,
@@ -198,8 +204,9 @@ class NotebookCollection(BaseCollection):
         ----------
         text : str, optional
             Free-text search term matched against block content.
-        project_id : ProjectId, optional
+        project_id : SearchProjectId, optional
             Scope the search to notebooks in the given project (format ``PRO...``).
+            The ``PRO`` prefix is stripped before the request is sent.
         sort_by : str, optional
             Field to sort on.
         order : OrderBy, optional
--- a/src/albert/collections/notebooks.py
+++ b/src/albert/collections/notebooks.py
@@ -1,5 +1,7 @@
 import mimetypes
+from collections.abc import Iterator
 from pathlib import Path, PurePosixPath
+from typing import Any
 
 from pydantic import TypeAdapter, validate_call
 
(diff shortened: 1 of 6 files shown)


sumedhb1995 (approved):
Thanks for putting this together, the change looks great overall. The search implementation follows the existing paginator pattern nicely, the NotebookSearchItem model is clean, and the test approach with poll_until is exactly right for index lag.

A couple of things I'd like addressed before this gets merged:

1. project_id handling: this PR strips the PRO prefix before sending (via SearchProjectId), but the sibling PRs for reports and product design send the full PRO id, and synthesis takes a raw string. Please confirm which form /api/v3/notebooks/search actually expects and align the convention across the set. As it stands, at least one of these PRs is sending the wrong thing.

2. The spec for this endpoint also lists a globalSearch parameter, which isn't exposed here. If leaving it out was a deliberate choice, a quick note in the PR description would help; otherwise it would be good to add, since tenant-wide search is the main use case this unblocks.

3. The new integration test hasn't been run against the live API yet, so please make sure CI is green before merging. The field aliases on NotebookSearchItem (Created/Updated in particular) are unverified against real responses until then.

Smaller, non-blocking note: the flaky-test fix in test_data_templates.py is unrelated to this change and would be cleaner as its own PR, but no harm done.

Thanks again for doing this, happy to re-review quickly once the above is sorted.

prasad-albert:
**Re: project_id, globalSearch, CI**

1. **project_id** — Confirmed against the backend: `/api/v3/notebooks/search` expects `projectId` without the `PRO` prefix. This PR uses `SearchProjectId`, consistent with inventory/property_data. Sibling PRs (#722, #723, #726) are now aligned on the same convention.

2. **globalSearch** — Deliberately left out for now. The OpenAPI spec documents the `globalSearch` query param but not the alternate grouped response shape (`{category, total, data}`). Noted as deferred in the PR description; waiting on the API team to confirm the contract before modeling it in the SDK.

3. **CI** — Green. Integration test passes against staging.

Non-blocking: the `test_data_templates.py` flaky-test fix was bundled for CI stability on this branch; happy to split if preferred.

PR #723, opened 2026-09-04
feat(reports): add search support

prasad-albert:
## What

Callers can now search saved analytical reports via `client.reports.search(...)` with free-text, project, report type, creator, facet, and contains filters, plus POST-only `source_field` / `additional_field` controls. Results are lightweight `ReportSearchItem` hits that hydrate into `FullAnalyticalReport`.

## Why

Implements SDK-101. Ask Albert and scripts need to resolve saved reports by name, project, or creator through OpenSearch without hand-rolling `session.post` calls.

## How

- POST-only search (matches `ProjectCollection.search` precedent); `AlbertPaginator` in OFFSET mode owns `offset`/`limit` internally, so the public method exposes only `max_items`.
- `ReportSearchItem.hydrate()` is overridden to call `get_full_report`, since `ReportCollection` has no `get_by_id` and the stock `HydrationMixin` path would fail.
- `parentId` is deliberately not exposed: the backend handlers ignore it despite the OpenAPI spec listing it.

## Testing

- [x] `uv run ruff format .`
- [ ] `uv run pytest tests/collections/test_reports.py -v -n 4` — new `test_search_reports` (scoped to seeded ids, `poll_until` for index lag, hydrate round-trip) added; integration run requires Albert credentials not available in the authoring environment, so it must run in CI.

## SDK Changes

Added `ReportCollection.search(...)` for OpenSearch-backed discovery of saved reports, returning `ReportSearchItem` results with `hydrate()` support.

Cake session: https://agents.ai.albertinventdev.com/sessions/813db603-8e47-445a-89b6-17709dd02c34

### DIFF
--- a/src/albert/collections/reports.py
+++ b/src/albert/collections/reports.py
@@ -1,11 +1,15 @@
+from collections.abc import Iterator
 from typing import Any
 
 from pydantic import validate_call
 
 from albert.collections.base import BaseCollection
+from albert.core.pagination import AlbertPaginator
 from albert.core.session import AlbertSession
-from albert.core.shared.identifiers import ReportId
-from albert.resources.reports import FullAnalyticalReport, ReportInfo
+from albert.core.shared.enums import OrderBy, PaginationMode
+from albert.core.shared.identifiers import ProjectId, ReportId
+from albert.core.utils import ensure_list
+from albert.resources.reports import FullAnalyticalReport, ReportInfo, ReportSearchItem
 
 
 class ReportCollection(BaseCollection):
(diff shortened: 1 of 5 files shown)


sumedhb1995 (approved):
This looks really good, thanks for the work. The hydrate() override routing to get_full_report is a smart catch since ReportCollection has no get_by_id, and going POST-only matches the ProjectCollection precedent for good reason.

Two things before this merges:

1. project_id is sent with the PRO prefix intact here, but the notebooks and worksheets PRs in this series strip it, and synthesis takes an unvalidated string. Please confirm what /api/v3/reports/search expects and get the whole set on one convention.

2. The integration test hasn't been run against the live API yet, so please wait for a green CI run before merging. The ReportSearchItem field aliases are unverified against real responses until that happens.

Great stuff otherwise, happy to re-review once these are sorted.

prasad-albert:
**Re: project_id, CI**

1. **project_id** — Updated to `SearchProjectId`. POST body sends `projectId` without the `PRO` prefix, aligned with the rest of the search PR set.

2. **CI** — Still red, and not because of `test_search_reports_project_id_filter` (removed; `project_id` coverage remains in `test_search_reports`). The failure is fixture setup: `seeded_reports` hits a 404 creating report type `ALB#RET42` in staging before the search test runs.

PR #722, opened 2026-09-04
feat(synthesis): add search support

prasad-albert:
## What

`client.synthesis.search(...)` is now available, wrapping `GET /api/v3/synthesis/search`. Callers can discover synthesis records via free text, chemical structure filters (SMILES, CAS, reactant/product names), project and inventory scope, and creator filters, with results returned as a lazily paginated iterator of lightweight `SynthesisSearchItem` hits (`id`, `name`).

## Why

Implements SDK-105. Ask Albert and scripts need to look up synthesis/reaction records by name, SMILES, CAS, or reactant/product chemistry without dropping to raw `session.get` calls.

## How

- Follows the established GET + `PaginationMode.OFFSET` search pattern (as in `CustomTemplatesCollection.search`); `max_items` is the only caller-facing pagination control, with `offset`/`limit` managed internally by `AlbertPaginator`.
- Array filters accept `str | list[str]` and are normalized with `ensure_list()`; camelCase wire params omit `None` values.
- `SynthesisSearchItem` stays minimal (`albertId` -> `id`, `name`) per the ticket; undeclared response fields are dropped by the default `extra="ignore"`. Full records are fetched via `get_by_id(id=hit.id)`.

## Testing

- [x] `uv run ruff format .`
- [x] `uv run ruff check . --fix`
- [ ] `uv run pytest tests/collections/test_synthesis.py -v -n 4` — blocked locally: no Albert credentials (`ALBERT_CLIENT_ID_SDK` / `ALBERT_CLIENT_SECRET_SDK` / `ALBERT_BASE_URL`) in this environment; test collects and fails only at client auth. Relies on CI.

New integration test `test_synthesis_search` (xdist group `projects`): creates a private synthesis on `seeded_notebooks[0]` named with `seed_prefix`, polls `search(text=seed_prefix)` via `poll_until` scoped to the created ID, asserts the hit, and confirms the hit ID round-trips through `get_by_id`. Cleanup runs in `try/finally`.

## SDK Changes

`SynthesisCollection.search(...)`: search synthesis records by free text, chemistry, project, or creator filters, returning lightweight `SynthesisSearchItem` hits.

Ticket: SDK-105
Cake session: https://agents.ai.albertinventdev.com/sessions/3eb2ffb8-1294-45a0-a643-6b3f6de9e2ca

PR #724, opened 2026-09-04
feat(worksheets): add search support

prasad-albert:
## What

Callers can now search formulations across worksheet grids with `client.worksheets.search(...)`, a lazily paginated iterator of lightweight `WorksheetSearchItem` results supporting free text plus project, sheet, tag, inventory, data template, creator, facet, and contains filters.

## Why

Ask Albert and script workflows need to discover formulations across projects without dropping to raw `session.get` calls (SDK-102). This unblocks cross-project formulation lookup in agent workflows.

## How

- `WorksheetCollection.search()` wraps the worksheet search endpoint with an OFFSET `AlbertPaginator`; `max_items` is the only caller-facing pagination control and `offset`/`limit` stay internal.
- New `WorksheetSearchItem` model (plus nested `WorksheetSearchInventoryLine` and `WorksheetSearchTag`) carries the lightweight formula hit shape; hits are not hydrated, so callers pass `hit.id` to `client.inventory.get_by_id(...)` when they need the full formula.
- Array filters accept a single value or a list via `ensure_list()`; hits deserialize directly with no collection binding.

## Testing

- [x] `uv run ruff format .`
- [x] `uv run pytest <relevant_test_file>.py -v`

Added `test_worksheet_search` to `tests/collections/test_worksheet.py` (`sheets` xdist group): scopes by `seed_prefix` and `project_id`, filters to `seeded_products` ids, and wraps the fetch in `poll_until`. Note: integration tests require `ALBERT_CLIENT_ID_SDK`, `ALBERT_CLIENT_SECRET_SDK`, and `ALBERT_BASE_URL`, which were not available in the authoring environment, so the new test was verified for collection/deserialization behavior locally but not executed against the live API; CI will exercise it.

## SDK Changes

Added `client.worksheets.search(...)` for discovering formulations across worksheets, returning lightweight `WorksheetSearchItem` results (id, name, inventory lines, tags).

### DIFF
--- a/src/albert/collections/worksheets.py
+++ b/src/albert/collections/worksheets.py
@@ -8,7 +8,7 @@ from albert.collections.custom_templates import CustomTemplatesCollection
 from albert.core.pagination import AlbertPaginator
 from albert.core.session import AlbertSession
 from albert.core.shared.enums import OrderBy, PaginationMode
-from albert.core.shared.identifiers import ProjectId
+from albert.core.shared.identifiers import ProjectId, SearchProjectId
 from albert.core.utils import ensure_list
 from albert.resources.acls import ACLContainer
 from albert.resources.custom_templates import CustomTemplate
@@ -147,7 +147,7 @@ class WorksheetCollection(BaseCollection):
         self,
         *,
         text: str | None = None,
-        project_id: ProjectId | None = None,
+        project_id: SearchProjectId | None = None,
         sheet_id: str | None = None,
         tags: str | list[str] | None = None,
         albert_id: str | list[str] | None = None,
@@ -185,8 +185,9 @@ class WorksheetCollection(BaseCollection):
         ----------
         text : str, optional
             Free-text query matched against formula fields.
-        project_id : ProjectId, optional
+        project_id : SearchProjectId, optional
             Scope the search to a Project (format ``PRO...``).
+            The ``PRO`` prefix is stripped before the request is sent.
         sheet_id : str, optional
             Filter to formulas on a given Sheet.
         tags : str or list[str], optional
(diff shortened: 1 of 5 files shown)


sumedhb1995 (approved):
Thanks for this one, the shape of the change looks great. The nested inventory line and tag models are a nice way to carry the formula hit data, and the filter set is comprehensive.

A few things to address before merging:

1. project_id handling: this PR strips the PRO prefix (via SearchProjectId), but the reports and product design PRs in this series send it intact. Please confirm what the worksheet search endpoint expects and align the convention across all of them.

2. The integration test hasn't been run against the live API yet, so please hold for green CI before merging. Worth double checking in particular that the ingredient lines on hits really come back keyed as id (no alias on WorksheetSearchInventoryLine.id), and that the endpoint accepts exclude_hidden serialized as True/False (that's the existing SDK convention for bools in query params, so probably fine, but unverified here).

Non-blocking: the poll_until refactors in test_data_columns.py and test_projects.py are unrelated to this change and would be cleaner in their own PR, but it's not a problem.

Thanks again, I'll happily re-review once the above is addressed.

prasad-albert:
**Re: project_id, CI**

1. **project_id** — Already uses `SearchProjectId` (strips `PRO` prefix). Backend worksheet search handler expects the bare numeric ID. Now aligned with #721, #722, #723, #726.

2. **CI** — Green. `exclude_hidden` serializes as `true`/`false` via the existing SDK bool convention (`json.dumps` in `_encode_query_params`), and the endpoint accepts it. Ingredient lines come back keyed as `id` with no alias — confirmed in integration run.

Non-blocking: the `poll_until` refactors in `test_data_columns.py` and `test_projects.py` were bundled for CI stability; can split if preferred.

PR #725, opened 2026-09-04
feat(teams): add search support

prasad-albert:
## What

Callers can now search teams via `client.teams.search(...)`, a lazily paginated iterator of lightweight `TeamSearchItem` hits backed by the team search endpoint. Each hit carries summary fields (`id`, `name`, creator/updater display names, timestamps, members) and can be upgraded to a fully populated `Team` via `hit.hydrate()`.

## Why

Resolves SDK-106. Ask Albert and scripts need to resolve teams by free text, status, or team ID without hand-rolling raw `session.post` calls. OpenSearch-backed `search` complements the existing exact-name `get_all` listing.

## How

- Follows the established `ProjectCollection.search` / `DataColumnCollection.search` pattern: POST-only, OFFSET-mode `AlbertPaginator`, `max_items` as the only caller-facing pagination control (`offset`/`limit` stay internal).
- `None` filters are omitted from the wire payload; array filters (`team_id`, `search_field`, `source_field`, `additional_field`) go through `ensure_list()`.
- New `TeamSearchItem` model uses `HydrationMixin[Team]`, so `hydrate()` delegates to `TeamCollection.get_by_id`.
- Purely additive: no existing signatures, return types, or models changed.

## Testing

- [x] `uv run ruff format .`
- [ ] `uv run pytest tests/collections/test_teams.py -v -n 4` — blocked locally: this environment has no `ALBERT_CLIENT_ID_SDK` / `ALBERT_CLIENT_SECRET_SDK` / `ALBERT_BASE_URL` credentials, so the integration suite cannot construct a client. New `test_search` follows the required pattern (scoped to `seed_prefix`, id-filtered to `seeded_team`, wrapped in `poll_until`, hydration assert) and is expected to run in CI.

## SDK Changes

Added `TeamCollection.search(...)` for free-text team search returning partial `TeamSearchItem` results with `hydrate()` support.

### DIFF
--- a/src/albert/collections/teams.py
+++ b/src/albert/collections/teams.py
@@ -1,15 +1,16 @@
 from collections.abc import Iterator
+from typing import Any
 
 from pydantic import validate_call
 
 from albert.collections.base import BaseCollection
 from albert.core.pagination import AlbertPaginator
 from albert.core.session import AlbertSession
-from albert.core.shared.enums import PaginationMode
+from albert.core.shared.enums import OrderBy, PaginationMode
 from albert.core.shared.identifiers import TeamId, UserId
 from albert.core.utils import ensure_list
 from albert.exceptions import AlbertException
-from albert.resources.teams import Team, TeamMember
+from albert.resources.teams import Team, TeamMember, TeamSearchItem
 from albert.resources.users import User
 
 
@@ -46,6 +47,8 @@ class TeamCollection(BaseCollection):
 
     Methods
     -------
+    search(...) -> Iterator[TeamSearchItem]
+        Search teams with free text and filters, returning partial items.
     get_all(name, exact_match, created_by, updated_by, user_id, max_items) -> Iterator[Team]
         Lists all teams with optional filters.
     get_by_id(id) -> Team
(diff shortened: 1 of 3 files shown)


sumedhb1995 (approved):
Thanks for putting this together, the change looks great and the hydrate flow through TeamCollection.get_by_id is exactly how it should work.

One blocker and one thing to verify before merging:

1. Likely bug: TeamSearchItem.members uses alias "user", but the Team model reads members from "Users" (capital U). If the search index returns the same key as the team endpoint, members will silently come back as None on every hit since undeclared fields are ignored. Related: TeamMember.id is required, so if search hits return partial member objects without an id, deserialization of the whole hit fails and the paginator will skip the team with only a log warning. Please verify the actual search response shape and adjust the alias (and consider making member id optional on the search item).

2. The integration test hasn't been run against the live API yet, so please wait for green CI before merging. That run is also what would confirm the point above.

Everything else looks solid. Thanks again, and ping me when it's ready for another look.

prasad-albert:
**Re: members alias, TeamMember.id**

1. **members alias** — Verified against `api-team` spec and indexer. Search and get-by-id use different wire keys by design:
   - Full `Team` resource → `Users` (capital U) — `Team.members` alias is correct.
   - Search hits (`TeamSearchResult`) → `user` (lowercase) — `TeamSearchItem.members` alias is correct.

   The OpenSearch stored fields and `api-indexer/postTeam.js` both index `user: [{id, name}]`. Using `Users` on search hits would silently drop members. No change needed.

2. **TeamMember.id required** — Search indexer always writes `{id, name}` per member; `role`/`fgc` is absent on search hits and `TeamMember.role` is already optional. CI green with live deserialization.

PR #726, opened 2026-09-04
feat(product_design): add search support

prasad-albert:
## What

`client.product_design` now exposes `search(...)`, wrapping the product design grid search endpoint so callers can discover formulations by free text, facets, and project / inventory filters. Search returns lightweight `ProductDesignSearchItem` hits (id, name, ingredient rows, tags) as a lazily paginated iterator. The existing unpack API is unchanged.

## Why

Ask Albert and scripts need to look up formulations on the product design grid without hand-rolling `session.get` calls. This unblocks product-grid formulation lookup for agents (SDK-107).

## How

- GET search with an OFFSET `AlbertPaginator`; callers control result volume via `max_items` only (no public `offset` / `limit`).
- The endpoint caps page size at 100, so the paginator params pin `limit` to that cap (same pattern as workflow search).
- Search hits are intentionally not hydrated; pass a hit's `id` to `client.inventory.get_by_id` for the full formula.
- Grid search (this change) and CAS-level unpack (`get_unpacked_products`) are separate APIs; the class docstring now calls that out.

## Testing

- [x] `uv run ruff format .`
- [ ] `uv run pytest tests/collections/test_product_design.py -v` (integration test added: `test_search` under the `sheets` xdist group, scoped to `seed_prefix` + `seeded_products` ids with `poll_until`; requires live API credentials, not run in this environment)

## SDK Changes

New: `client.product_design.search(...)` and `ProductDesignSearchItem` for formulation discovery on the product design grid.

---

Cake session: https://agents.ai.albertinventdev.com/sessions/593afeff-f69a-4346-8a19-c307607f2c88

### DIFF
--- a/src/albert/collections/product_design.py
+++ b/src/albert/collections/product_design.py
@@ -1,25 +1,38 @@
-from typing import Literal
+from collections.abc import Iterator
+from typing import Any, Literal
 
 from pydantic import validate_call
 
 from albert.collections.base import BaseCollection
+from albert.core.pagination import AlbertPaginator
 from albert.core.session import AlbertSession
-from albert.core.shared.identifiers import InventoryId
-from albert.resources.product_design import UnpackedProductDesign
+from albert.core.shared.enums import OrderBy, PaginationMode
+from albert.core.shared.identifiers import InventoryId, ProjectId
+from albert.core.utils import ensure_list
+from albert.resources.product_design import ProductDesignSearchItem, UnpackedProductDesign
+
+_SEARCH_PAGE_SIZE = 100  # maximum page size accepted by the product design search endpoint
 
 
 class ProductDesignCollection(BaseCollection):
-    """Unpack formulated products into their full substance-level composition.
+    """Search the product design grid and unpack formulated products.
 
-    "Product design" here refers to unpacking (flattening) a formulation inventory
-    item into its full substance-level composition. An unpacked product resolves a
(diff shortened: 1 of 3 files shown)


sumedhb1995 (approved):
This looks great, thanks for the work on it. Pinning the page size to the endpoint's 100 limit is a thoughtful detail, and the updated class docstring clarifying search vs unpack is a genuine improvement.

Two things before merging:

1. project_id is sent with the PRO prefix intact here, while the notebooks and worksheets PRs in this series strip it. Please confirm what /api/v3/productdesign/search expects and get the set aligned on one convention.

2. The integration test hasn't been run against the live API yet, so please hold merge until CI is green. Worth confirming the product_grid bool is accepted as True/False in the query string (that matches existing SDK convention for bool params, so likely fine, just unverified).

Non-blocking note: same as the worksheets PR, the ingredient line model assumes the wire key is id with no alias, so keep an eye on that when CI runs.

Thanks again, happy to re-review once these are sorted.

prasad-albert:
**Re: project_id, CI**

1. **project_id** — Updated to `SearchProjectId`. Backend product design search expects `projectId` without the `PRO` prefix, consistent with notebooks/worksheets/synthesis/reports.

2. **CI** — Green. `product_grid` bool serializes as `true`/`false` per SDK convention and the endpoint accepts it. Ingredient line `id` wire key confirmed in integration run.

PR #728, opened 2026-09-08
feat(intervals): support increased intervals

prasad-albert:
## What

Callers on increased-intervals tenants can create intervalized Property tasks, configure and persist block exclusion/inclusion rules and overrides, automatically generate child-workflow combinations via background worker jobs, and inspect combination items with barcode preservation.

## Why

Flag-ON tenants replace legacy positional ROW intervals with denormalized child workflows, lift intervalized-parameter caps, and require client-side Cartesian product generation, rule evaluation, and worker job orchestration.

## How

- `create_with_combinations` orchestrates bulk workflow creation, task creation, rules persistence, per-block combination generation, and enforces a mandatory 10-second gap between worker jobs.
- `generate_block_combinations` evaluates combination rules client-side, signs and uploads the S3 payload, posts the `createChildWorkflows` worker job, and polls to terminal state.
- `set_block_rules` persists rules and automatically re-evaluates and regenerates child-workflow combinations on Albert Invent by default, passing the current workflow as `old_workflow_id` so unchanged barcodes are preserved and obsolete combinations are voided.
- Value-based helpers on `Workflow` (`build_rule`, `build_rule_condition`, `build_override`) allow callers to define rules and overrides using human-readable parameter names and values rather than manually looking up workflow-local IDs (`PRG...`, `PRM...`, `ROW...`).
- Helpers use `@validate_call` with keyword-only arguments (`*`), and support both `Condition` NamedTuples and plain tuples (e.g. `("Temperature", ">=", 90)`).
- Dual-mode evaluation (`intervalsStartFrom: 'all' | 'none'`) in `generate_interval_combinations` accurately models frontend parity for JS numeric coercion, override precedence, and condition matching.
- Partial failures in `create_with_combinations` raise `CombinationGenerationError`, carrying the created task instance and failed block IDs for targeted retries.
- All increased-intervals methods and models carry `(🧪 Beta)` badges and feature flag warning callouts.

## Testing

- [x] `uv run ruff format .`
- [x] `uv run pytest tests/unit/test_tasks_combinations.py tests/unit/test_interval_combinations_engine.py -v`
- [x] `uv run pytest tests/resources/test_interval_combinations.py -v`
- [x] `uv run pytest tests/core/shared/test_identifiers.py -v`
- [x] `uv run mkdocs build --strict`

## SDK Changes

- Added `TaskCollection.create_with_combinations`, `TaskCollection.generate_block_combinations`, `TaskCollection.get_block_combinations`, `TaskCollection.get_block_rules`, and `TaskCollection.set_block_rules` (🧪 Beta).
- Added models `IntervalCombinationItem`, `RuleOperator`, `OverrideAction`, `Condition`, `RuleCondition`, `ExclusionRule`, `CombinationOverride`, `BlockRules`, and S3 payload models (🧪 Beta).
- Added `Workflow.build_rule_condition`, `Workflow.build_rule`, and `Workflow.build_override` to construct rules and overrides using parameter names and values with keyword-only arguments (🧪 Beta).
- Added `Workflow.get_override_key` to build canonical override keys from parameter values (🧪 Beta).
- Added `CombinationGenerationError` for resilient error handling and partial failure recovery.
- Added `poll_worker_job` in `albert.utils.worker_jobs` for background worker job polling.

### DIFF
--- a/src/albert/resources/interval_combinations.py
+++ b/src/albert/resources/interval_combinations.py
@@ -86,11 +86,17 @@ class OverrideAction(str, Enum):
 
 
 class Condition(NamedTuple):
-    """Specification for a rule condition before resolving workflow-local IDs.
+    """Specification for a rule condition before resolving workflow-local IDs (🧪 Beta).
 
     Used with [`Workflow.build_rule`][albert.resources.workflows.Workflow.build_rule] to specify
     criteria using parameter names or short names. Can be instantiated as a named tuple or
     passed as a plain tuple (e.g. ``("Temperature", ">=", 90)``).
+
+    !!! warning "Beta Feature!"
+        Increased intervals combination support is currently in beta and behind a platform
+        feature flag. Please do not use in production or without explicit guidance from
+        Albert. You might otherwise have a bad experience. This feature currently falls
+        outside of the Albert support contract, but we'd love your feedback!
     """
 
     parameter: str
--- a/src/albert/resources/interval_combinations.py
+++ b/src/albert/resources/interval_combinations.py
@@ -87,7 +87,7 @@ class RuleCondition(BaseAlbertModel):
     operator: RuleOperator
     """Comparison operator used to evaluate this condition."""
 
-    value: str | float | int
+    value: str | float | int | None = Field(default=None, alias="value")
     """Threshold value to compare against."""
 
     unit_id: str | None = Field(default=None, alias="unitId")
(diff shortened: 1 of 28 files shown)


src/albert/resources/interval_combinations.py
```diff
@@ -0,0 +1,322 @@
...
+    """
+
+    parameter_group_id: ParameterGroupId | DataTemplateId | str = Field(alias="prgId")
+    """Parameter group or data template ID (format ``PRG...`` or ``DAT...``)."""
+
+    parameter_id: ParameterId = Field(alias="prmId")
+    """Parameter ID (format ``PRM...``)."""
+
+    row_id: RowId | None = Field(default=None, alias="rowId")
+    """Parameter row ID within the workflow (format ``ROW...``)."""
```
lkubie:
```suggestion
    """Parameter row ID within the workflow (format ``ROW...``). Required to distinguish parameters because a single Parameter ID can be repeated within a parameter group."""
```

src/albert/resources/workflows.py
```diff
@@ -37,11 +46,20 @@ class IntervalParameter(BaseAlbertModel):
     interval_id: IntervalId | None = Field(default=None)
     """The row ID of this single interval value (e.g. ``"ROW1"``). These are the building blocks that [`get_interval_id`][albert.resources.workflows.Workflow.get_interval_id] joins with ``X`` to form a composite interval ID."""
```
lkubie:
```suggestion
    """The row ID of this single interval value (e.g. ``"ROW1"``). Parameters can be repeated within a Parameter Group, so the Parameter ID is not unique enough to identify which parameter is being varied. So, these key become the building blocks that [`get_interval_id`][albert.resources.workflows.Workflow.get_interval_id] joins with ``X`` to form a composite interval ID."""
```

lkubie (approved):
Doc Nits but LGTM

PR #735, opened 2026-09-15
feat(parameters): add SpecialParameterType enum and special_type field to ParameterValue and TargetParameter

anubhavd-albert:
## Summary

- Defines `SpecialParameterType` enum in `resources/parameters.py` with members `RawMaterials`, `Consumables`, `Equipment` -- the canonical source of truth for special parameter subtypes, verified against dev (PROMO68815 via RET48/RET52).
- Adds `special_type: SpecialParameterType | None` as a writable optional field on `ParameterValue` and `TargetParameter`.
- Field is absent from `Parameter`: the platform's parameter API does not return the subtype (confirmed by inspecting the raw `GET /parameters/{id}` response and `api-parameter`'s `api.yaml`). The field exists only in Zeus report data.

Closes ML-1611. Unblocks ML-1612.

## Test plan

- [ ] Live dev test run against PROMO68815: `SpecialParameterType` enum members match RET48/RET52 `Special Parameter Type` column values (`RawMaterials`, `Consumables`, `Equipment`).
- [ ] Responses missing `special_type` deserialise without error (older tenants and cached payloads).
- [ ] `ParameterValue` and `TargetParameter` accept `special_type` on construction and round-trip correctly.


src/albert/resources/parameters.py
```diff
@@ -6,6 +6,34 @@
 from albert.core.shared.types import MetadataItem
 
 
+class SpecialParameterType(str, Enum):
+    """The specific entity type referenced by a Special
+    [`Parameter`][albert.resources.parameters.Parameter].
+
+    Not returned by the platform parameter API (use Zeus reports RET48/RET52 to
```
sbernasek, line 13: Zeus is a private API whereas the SDK is public; we should try to avoid referencing it in our docstrings as they're visible external to Albert

sbernasek:
Please take a pass to strip zeus references from the docstrings, but other than that lgtm!

sbernasek (approved):

PR #736, opened 2026-09-16
feat(chats): add permissionActions to ChatMessage (AI-1832)

sumedhb1995:
## What

`ChatMessage` gains a `permissionActions` field, so a chat message can carry a batch of permission decisions instead of at most one.

## Why

AI-1832 replaces Ask Albert's single all-writes grant with per-collection-group permissions. A turn can now raise several permission cards, the user answers each, and the frontend sends every decision in one message.

This SDK is a **hard sequencing dependency for that work, and it fails silently**. The ReAct worker reads and writes chat rows through this model, and pydantic v2 drops unknown fields without complaint. Until this field exists, the batch is written and read back as nothing: "Allow for this chat" does not survive a turn boundary and every gated write re-prompts, with no error anywhere to explain why.

## How

- `permission_actions: list[PermissionAction] | None`, serialized as `permissionActions` to match the api-chat contract.
- Each action carries a `permissionKey` (`"<group>.<write|delete>"`) so a grant is scoped to the collection group the user actually approved, rather than to all writes.
- Additive and optional. Older payloads round-trip unchanged, and the existing single-action field is untouched.

## Testing

`pytest tests/resources/test_chats.py` — covers the new field's round-trip in both directions, that omitting it is still valid, and that the camelCase alias is what goes on the wire.

Consumers cannot use this until it is released. albert-python uses release-please, so the release happens on merge to main; the albert-ai worker currently floors at `albert>=1.53.0,<2` with a runtime probe that fails closed, and its floor should be raised to the release carrying this field once it lands.


prasad-albert:
> Code Review
> 
> Issues Found
> 
> * src/albert/resources/chats.py:302 — The edited permission_action attribute docstring ends with a dangling "See Also --------" fragment (nothing follows it). Attribute docstrings become the field JSON-schema description, so this stray NumPy-section marker leaks into generated docs/tooling output. Drop the trailing "See Also --------" since the line was already being touched.
> 
> Summary 1 minor issue found. No correctness or runtime bugs; the new permission_actions field, alias, type hints, and round-trip tests are all correct.

@claude Fix the attribute docstring

prasad-albert (approved):
LGTM

PR #740, opened 2026-09-17
feat(client): support caller-supplied default headers on both sessions

sumedhb1995:
## Why

Ask Albert stamps turn context onto its platform calls so an activity record can be traced back to the conversation that produced it. The platform reads that context from an `x-s2s-metadata` request header on the resource call, not from the token, so the SDK client needs to send a fixed header on every request it makes.

The SDK had no supported way to do that. `AlbertSession` subclasses `requests.Session`, so the sync path has a usable `.headers`. `AsyncAlbertSession` is a plain wrapper whose only header surface is the private `_client` attribute on the inner `httpx.AsyncClient`, and the async consumer in `albert-ai` has been mutating that directly. It works, but it is outside the public contract: a refactor of that session would break header forwarding with no import error and no failing test on either side.

## What

Adds a keyword-only `headers: dict[str, str] | None = None` to:

- `AlbertSession.__init__` and `AsyncAlbertSession.__init__`
- `Albert.__init__` / `AsyncAlbert.__init__`
- `Albert.from_token`, `Albert.from_sso`, `Albert.from_client_credentials`
- `AsyncAlbert.from_token`, `AsyncAlbert.from_client_credentials`

Headers given there apply to every request the client makes.

## Precedence

Caller headers merge **over** the SDK defaults, so `Content-Type`, `Accept`, and `User-Agent` can be overridden deliberately.

`Authorization` is the exception and cannot be displaced. Both transports set it per request from the session's own credentials: on the async side that lands after the client defaults in httpx's merge order, and on the sync side the per-request dict wins over the session dict. `test_authorization_cannot_be_overridden_by_default_headers` pins this, since a caller-settable auth header would be a real footgun.

## Compatibility

Purely additive and keyword-only. No existing call site changes, no behavior changes when the argument is omitted.

## Testing

New `tests/core/test_default_headers.py`, 7 cases: both sessions apply the headers, SDK defaults survive, omitting the argument changes nothing, both `from_token` factories pass through, the header is on the wire across repeated requests (`MockTransport`, not just on the client object), and the `Authorization` guarantee above. Full offline suite (`tests/core`, `tests/unit`, `tests/test_exceptions.py`) passes at 105.

Versioning is release-please's; the `feat` prefix takes this to a minor.

## Follow-up

Once this releases, the `albert-ai` side switches both its `x-s2s-metadata` and its existing `x-external-user-id` binding off `session._client.headers` and onto this argument.


tests/core/test_default_headers.py
```diff
@@ -0,0 +1,94 @@
+"""Caller-supplied default headers on the sync and async sessions.
+
+The motivating case is audit metadata: Ask Albert stamps turn context onto
```
prasad-albert: Can we remove the AA reference here?

src/albert/core/async_session.py
```diff
@@ -44,6 +51,7 @@ def __init__(
                 "Content-Type": "application/json",
                 "Accept": "application/json",
                 "User-Agent": f"albert-SDK V.{albert.__version__}",
+                **(headers or {}),
```
prasad-albert: Dict unpack is case-sensitive. If a caller passes lowercase headers (e.g., `{"content-type": "..."}`), `httpx` sends both headers comma-joined on the wire, which can break backend JSON parsing. Using `httpx.Headers.update()` ensures case-insensitive replacement, matching `AlbertSession`:

```python
default_headers = httpx.Headers(
    {
        "Content-Type": "application/json",
        "Accept": "application/json",
        "User-Agent": f"albert-SDK V.{albert.__version__}",
    }
)
if headers:
    default_headers.update(headers)

self._client = httpx.AsyncClient(
    base_url=base_url,
    headers=default_headers,
)
```

src/albert/core/session.py
```diff
@@ -54,6 +61,8 @@ def __init__(
                 "User-Agent": f"albert-SDK V.{albert.__version__}",
             }
         )
+        if headers:
+            self.headers.update(headers)
```
prasad-albert, line 68: Minor consistency & hardening note: in `AlbertSession.request()` (line 97), setting `Authorization` *after* merging `kwargs` aligns with `AsyncAlbertSession` and ensures a per-request header cannot override the session auth token:

```python
headers = kwargs.pop("headers", None) or {}
headers["Authorization"] = f"Bearer {self._access_token}"
kwargs["headers"] = headers
```

src/albert/client.py
```diff
@@ -124,6 +124,7 @@ def __init__(
         auth_manager: AlbertClientCredentials | AlbertSSOClient | None = None,
         retries: int | None = None,
         timeout: float | tuple[float, float] | None = None,
+        headers: dict[str, str] | None = None,
```
prasad-albert: Prefer `Mapping[str, str] | None = None` (from `collections.abc`) over `dict[str, str] | None`. Since `dict` is invariant in type checkers, this lets callers pass `httpx.Headers` or `CaseInsensitiveDict` without type errors or explicit `dict()` casts.

prasad-albert (approved):

PR #741, opened 2026-09-17
fix(reports): strip ALB# qualifier from report IDs in URL paths

prasad-albert:
## What

`client.reports.get_report` (and its analytics/datascience wrappers) and `client.report_templates.get_by_id` now accept fully qualified report IDs such as `ALB#RET40` — exactly the form `report_template_get_all` returns as `albertId` — and normalize them to the bare form before building the request URL.

## Why

Both methods interpolated the caller-supplied ID straight into the URL path. Because `#` starts a URL fragment, a qualified ID like `ALB#RET40` was truncated to `GET /api/v3/reports/analytics/ALB`, which the platform answers with 404 "No Record Found". Round-tripping a listed template into a report run therefore failed, even though the `get_report` docstring advertises `"ALB#RET51"` as valid input. This is the root cause of AI-1679 (TEN5 report pull errors). Fixes SDK-109.

## How

- Normalize with `id.rsplit("#", 1)[-1]` at the two interpolation sites (`reports.py::get_report`, `report_templates.py::get_by_id`). Bare IDs contain no `#` and pass through unchanged, so existing callers are unaffected.
- Chose stripping over URL-encoding the segment: the platform is known to accept the bare `RET…` form (the ReAct worker's interim `TODO(albert-sdk)(SDK-109)` workaround in `svc_ai_react_worker` strips the qualifier the same way and is deployed working behavior), while acceptance of an encoded `ALB%23RET40` is unverified.
- Docstrings on both parameters now state that fully qualified IDs are normalized automatically.
- No version bump — release-please owns versioning on merge.

## Testing

- Offline verification of request paths with a capture session: `ALB#RET51` → `/api/v3/reports/datascience/RET51`, `ALB#RET40` → `/api/v3/reporttemplates/RET40`, and bare IDs (`RET22`, `RET40`) pass through unchanged.
- New integration regression test `test_report_template_get_by_id_accepts_fully_qualified_id` round-trips a qualified `albertId` from `get_all` into `get_by_id` (skips at runtime if no qualified templates exist in the test environment, matching the existing skip in that file).
- `ruff format` / `ruff check` clean; offline suite (`tests/unit`, 30 tests) passes; the new test collects under pytest. Live integration runs require `ALBERT_CLIENT_ID_SDK` / `ALBERT_CLIENT_SECRET_SDK` / `ALBERT_BASE_URL` and were not executed here.

Follow-up (separate repos, after this SDK fix ships): remove the interim sanitizer in `svc_ai_react_worker/agent/sdk_tools/factory/runtime.py` and the bare-form instruction in the reporting bundle, per SDK-109.


prakashchandrahv (approved):
Removing residue items till `#`  is the only change here, looks good

PR #742, opened 2026-09-17
ci(lambda-layer): build from tag, multi-region, idempotent publish, docs

lkubie:
Suggestions on top of the lambda layer CircleCI work:

- Take the SDK version from the release tag instead of a typed parameter; the pipeline must run on a tag and validate_release_tag.py confirms it
- Build the layer from a wheel built from the checkout (uv build --wheel) rather than installing albert==X from PyPI, and assert the imported version inside the Lambda image
- Default runtimes to 3.11,3.12,3.13,3.14 (3.10 deprecates on Lambda 2026-10-31; 3.13 and 3.14 are GA)
- lambda_region becomes lambda_regions (comma list, default us-west-2); publish loops over regions with one zip
- Skip publishing in a region that already has a layer version for this SDK version, matched on a stable description prefix
- Layer name albert-python-py312-x86_64, zip albert-python-<ver>-py3.12-<arch>.zip
- Add docs/lambda.md explaining which layer to use and how to attach it


docs/lambda.md
```diff
@@ -0,0 +1,81 @@
...
+
+```text
+albert-python 1.19.0 | python3.12 | x86_64 | 2026-09-17T14:02:00Z | sha=abc1234
+```
+
+The ARNs for each release are listed in the [GitHub release notes](https://github.com/albert-labs/albert-python/releases). Pin your function to a specific layer version ARN and upgrade deliberately, the same way you would pin a package version.
+
+## Attaching the layer
+
+=== "AWS CLI"
```
prasad-albert, line 46: The documentation mentions:
> *"The ARNs for each release are listed in the [GitHub release notes](https://github.com/albert-labs/albert-python/releases). Pin your function to a specific layer version ARN..."*

Right now, there is a gap between the docs and the pipeline: the publish script only prints the generated `LAYER_ARN` to `stdout` within individual CircleCI matrix jobs. Nothing captures, aggregates, or posts these ARNs to the GitHub Release notes (which are created separately via `release-please`).

As it stands, external users have no way to discover what the published ARNs or version numbers actually are for a given release.

.circleci/scripts/publish-lambda-layer.sh
```diff
@@ -118,45 +119,70 @@ GIT_SHA="${GIT_SHA:-unknown}"
...
+IFS=',' read -ra REGION_LIST <<<"${REGIONS}"
+
+for REGION in "${REGION_LIST[@]}"; do
+  REGION="${REGION// /}"
+  [[ -z "$REGION" ]] && continue
+  echo "== ${REGION}: ${LAYER_NAME}"
+
+  # Idempotent: if a layer version for this SDK version already exists, reuse it.
+  EXISTING_ARN="$(
+    aws lambda list-layer-versions \
```
prasad-albert, line 136: By default, `aws lambda list-layer-versions` returns at most 50 items in a single response unless paginated with continuation tokens.

Once a layer name accumulates more than 50 versions over its lifecycle, older layer versions will be truncated from this query. As a result, the prefix match will fail to see existing versions beyond the first 50, causing the idempotency check to break down over time.

.circleci/config.yml
```diff
@@ -6,18 +6,15 @@ parameters:
   lambda_layer:
     type: boolean
     default: false
```
prasad-albert, line 8: Because `lambda_layer` defaults to `false`, when a release tag is pushed by `release-please`, only the standard `release` workflow (PyPI publish + docs deploy) will run. The `lambda_layer_matrix` workflow will not trigger automatically on tag creation, meaning layer builds still require a separate manual pipeline trigger in CircleCI with `lambda_layer: true`. Is that intended for now, or should the matrix be tied into the release workflow?

.circleci/scripts/publish-lambda-layer.sh
lkubie: You're right. AWS CLI v2 does auto-paginate this call (botocore has a ListLayerVersions paginator and --query runs on the merged result), but I'd rather not depend on that. I changed to check to now walk NextMarker pages explicitly with --no-paginate, stopping at the first match. 
LMK what you think about this approach.
I Tested against a stub aws that hides the matching version at the end of page 3. While testing I also found that a failed list call was previously swallowed inside the command substitution and fell through to a publish; it now aborts correctly.

docs/lambda.md
lkubie: Good catch 🤦‍♀️

 Each matrix job now writes the ARNs it published to a workspace manifest, and a final lambda_layer_release_notes job (requires all matrix jobs) merges them into a table, stores it as a CircleCI artifact, and PATCHes it into the GitHub release body for $CIRCLE_TAG between  markers, so **re-runs replace rather than duplicate**. Stdlib only. It authenticates with GITHUB_TOKEN_ALBERT_PYTHON from the dev context.

.circleci/config.yml
lkubie: Not intended, thanks for pushing on it. lambda_layer_setup now also runs in the release workflow after Publish to PyPI on every v* tag, so layers ship automatically with each release. The parameter and the manual lambda_layer_matrix workflow remain for ad hoc branch runs. I sequenced it after the PyPI publish so a failed release does not produce layers for a version that never shipped; happy to loosen that if you'd rather they be independent.

prasad-albert (commented):
Overall this is looking great! The prior review feedback around release notes ARN discovery, `list-layer-versions` pagination, and release tag workflow integration is thoroughly addressed and works cleanly. Just noticed two minor doc/config consistency items inline to adjust before merging.

docs/lambda.md
```diff
@@ -0,0 +1,81 @@
...
+| `albert-python-py311-x86_64` | python3.11 | x86_64 |
+| `albert-python-py311-arm64` | python3.11 | arm64 |
+| `albert-python-py312-x86_64` | python3.12 | x86_64 |
+| `albert-python-py312-arm64` | python3.12 | arm64 |
+| `albert-python-py313-x86_64` | python3.13 | x86_64 |
+| `albert-python-py313-arm64` | python3.13 | arm64 |
+| `albert-python-py314-x86_64` | python3.14 | x86_64 |
+| `albert-python-py314-arm64` | python3.14 | arm64 |
+
+Layers are currently published in `us-west-2`. Lambda layers are region scoped, so your function must be in the same region as the layer.
```
prasad-albert: Thanks for adding multi-region publishing support! With the recent additions of `us-east-1`, `eu-central-1`, and `eu-west-1` to the default regions in `.circleci/config.yml`, this section still mentions only `us-west-2`. Could we update this to list all four published regions so users in other regions know layers are available?

```suggestion
Layers are published in `us-west-2`, `us-east-1`, `eu-central-1`, and `eu-west-1`. Lambda layers are region scoped, so your function must be in the same region as the layer.
```

.circleci/scripts/generate_lambda_matrix.py
```diff
@@ -91,21 +143,18 @@ def generate(
...
+        '    default: "3.11,3.12,3.13,3.14"',
         "  lambda_archs:",
         "    type: string",
-        "    default: \"x86_64,arm64\"",
-        "  lambda_region:",
+        '    default: "x86_64,arm64"',
+        "  lambda_regions:",
         "    type: string",
-        "    default: \"us-east-1\"",
+        '    default: "us-west-2"',
```
prasad-albert: Could we update the fallback default for `lambda_regions` here to match the 4 regions defined in `.circleci/config.yml` (`"us-west-2,us-east-1,eu-central-1,eu-west-1"`)? That way, if this continuation config is ever inspected or evaluated standalone with defaults, it stays in sync with the primary workflow configuration.

```suggestion
        '    default: "us-west-2,us-east-1,eu-central-1,eu-west-1"',
```

prasad-albert (approved):

